PR #39, opened 2021-12-17
Nmr 4587 version number in a single file

vfiack:


pom.xml
```diff
@@ -4,10 +4,11 @@
 
     <groupId>org.nmrfx</groupId>
     <artifactId>nmrfx</artifactId>
-    <version>11.2.1</version>
+    <version>${revision}</version>
     <packaging>pom</packaging>
 
     <properties>
+        <revision>11.2.1</revision>
```
vfiack, line 11: This is the single place where you need to define the version.

I had to move the bom declaration in all modules to avoid having to redeclare the version in the bom as well.

PR #42, opened 2021-12-20
NMR-4576: JMX connector as a plugin

vfiack:


brucejohnson (approved):
Looks good other than the  printStackTrace that sonar highlighted.

vfiack:
> Looks good other than the printStackTrace that sonar highlighted.

Yes. I guess the next step is really decide which logging API we'll use, then start using it instead of all these System.out calls. 

On spinlab's side, we opted with java.util.logging (because it was there, not after comparing with other solutions) and added a few utility classes to simplify its usage. It's not perfect but at least doesn't come with an additional dependency.

PR #54, opened 2022-01-05
NMR-4493 Import jcamp-parser as dependency.

darbymcgraw:
Separated the dependency import as it's own PR to isolate the import in case it causes errors. This needs to wait for the spelling error to be fixed in the jcamp-parser module-info.java file to be accurate.

### DIFF
--- a/nmrfx-bom/pom.xml
+++ b/nmrfx-bom/pom.xml
@@ -22,7 +22,7 @@
             <dependency>
                 <groupId>com.nanalysis</groupId>
                 <artifactId>jcamp-parser</artifactId>
-                <version>1.0</version>
+                <version>1.0.1</version>
             </dependency>
             <dependency>
                 <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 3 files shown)


nmrfx-bom/pom.xml
```diff
@@ -19,6 +19,11 @@
 
     <dependencyManagement>
         <dependencies>
+            <dependency>
+                <groupId>com.nanalysis</groupId>
+                <artifactId>jcamp-parser</artifactId>
+                <version>1.0</version>
```
vfiack, line 25: This needs to change to get the fix about the module name:
```suggestion
                <version>1.0.1</version>
```

vfiack: Related PR: 
https://github.com/nanalysis/jcamp-parser/pull/3

For reference, I put the built artifact there:
https://github.com/nanalysis/maven-repository/tree/public/com/nanalysis/jcamp-parser/1.0.1

Which is then referenced as an additional maven repository in our projects. We should not replace released versions, only add new ones. This is open sourced, other people could depend on it.

vfiack (approved):

PR #56, opened 2022-01-07
NMR-4648: expose a way to generate an automatic script from JMX

vfiack:


nmrfx-jmx-connector/src/main/java/org/nmrfx/jmx/mbeans/Analyst.java
```diff
@@ -49,6 +49,17 @@ public void setWindowOnFront() {
...
     }
 
+    @Override
+    public void generateAutoScript() {
+        ConsoleUtil.runOnFxThread(() -> {
+            // getting the chart processor must be done in fx thread, because otherwise, a client calling
+            // open() then generateAutoScript() could try to access the chart processor before its creation.
+            ChartProcessor chartProcessor = FXMLController.getActiveController().getChartProcessor();
+            String script = chartProcessor.getGenScript(false);
+            FXMLController.getActiveController().getProcessorController(true).parseScript(script);
```
vfiack, line 59: @brucejohnson this is working, but is there a better way to do it?

brucejohnson: That's probably the best way to do it currently. 
The arrayed parameter is used for data that is pseudo-2D and you only want the FT etc. to happen in the direct dimension.  This could used for data like T1 relaxation.   I'm not sure how one would recognize that in RS2D files.  It looks like relaxation experiments have an array of tau values.

vfiack: for spinit viewer, we decide whether to display it as a contour (=2D) or stack plot (=pseudo 2D) depending on the number of rows. I can do something similar here.

darbymcgraw:
I've reviewed & tested but I don't entirely know what the goal is. I tested with a T1 experiment and I'm not getting any errors, but it doesn't seem like a processing script is running?

vfiack:
> I've reviewed & tested but I don't entirely know what the goal is. I tested with a T1 experiment and I'm not getting any errors, but it doesn't seem like a processing script is running?

This is normal. It allows calling the processing from the outside (either spinit or a JMX console), but if no one calls it, then nothing should be done.

darbymcgraw (approved):

PR #61, opened 2022-01-27
Nmr 4668 simpler spinitexport

vfiack:
small improvements, could still be better.

### DIFF
--- a/nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/ChartProcessor.java
+++ b/nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/ChartProcessor.java
@@ -62,6 +62,7 @@ import org.nmrfx.utils.GUIUtils;
  * @author brucejohnson
  */
 public class ChartProcessor {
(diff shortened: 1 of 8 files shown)


nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/ChartProcessor.java
```diff
@@ -1254,6 +1250,8 @@ void updateCounter() {
 
     public void setData(NMRData data, boolean clearOps) {
         setNMRData(data);
+        setDatasetType(data.getPreferredDatasetType());
```
vfiack, line 1253: @brucejohnson is this the correct location to do this?

It seems to work when loading a FID, I couldn't find whether it is also called when opening a dataset (but I think it isn't).

darbymcgraw: I don't see this getting called when opening a dataset.

darbymcgraw (approved):
Looks good to me. Was able to process with a spinit setting.

PR #62, opened 2022-01-27
Nmr 4682 notification on data save

vfiack:
includes https://github.com/nanalysis/nmrfx/pull/61 and https://github.com/nanalysis/nmrfx/pull/59 but needed to merge master, otherwise it would miss script generation from spinit and break interactions from there.

Only to be merged once the two others PR are.


darbymcgraw (approved):
Looks good to me

brucejohnson (approved):

PR #66, opened 2022-02-11
remove unused methods related to Dataset

vfiack:
also remove DatasetListener, which had no implementation.


brucejohnson (approved):
I expect that someday I may go back and find some of this old code and reuse it, but I'm fine with cleaning it up now.

PR #67, opened 2022-02-11
NMR-4709: remove NMRFxServer, replaced by JMX connector

vfiack:
I thought it was already removed, but found it again while replacing loggers.


brucejohnson:
Sorry, I had worked on this, but hadn't pushed it yet.  But you've got the same stuff.
Except, shouldn't netty be removed from the pom.xml files

vfiack:
> Except, shouldn't netty be removed from the pom.xml files

Good catch, thanks!

brucejohnson (approved):

PR #70, opened 2022-02-17
NMR-4798: logging path defined by conditionals

vfiack:


brucejohnson:
I tested on my Mac and it worked fine there (created log file in Application Support)

brucejohnson (approved):

PR #72, opened 2022-02-18
Nmr 4493 rewrite jcamp data from scratch

vfiack:
Ready for review and tests.


nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
```diff
@@ -17,987 +17,569 @@
...
-            return null;
-        } else {
-            return parMap.get(parname);
-        }
+    public int getNVectors() {
+        //XXX To confirm, looks like it is number of 1D rows in total
+        // so 1 if 1D data, and number of 2D if 2D data.
+        // no more than 2D data are supported in JCamp
+
+        //XXX should we try to get it from page content instead?
```
vfiack, line 143: Not sure I completely understood what getNVectors() is supposed to return.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
```diff
@@ -17,987 +17,569 @@
...
-            return Double.parseDouble(parMap.get(parname));
-        }
+    public int getNPoints() {
+        // XXX original JCAMPData always returned 0 (np / 2, with np never initialized). Isn't it a bug?
+
+        // XXX always /2, or only if complex data?
+        // XXX should we try to get it from page content instead?
+        return block.optional($TD, 0)
+                .map(r -> r.getInt() / 2)
+                .orElse(0);
```
vfiack, line 157: This is dubious as well. I'm pretty sure the original had a bug with `np` never initialized.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
```diff
@@ -17,987 +17,569 @@
...
     }
 
     @Override
-    public int getNPoints() {  // points per vector
-        return np / 2;
+    public void setSize(int dim, int size) {
+        //TODO implement me?
+        // Don't see how we would need this as we are accessing the size directly
+        System.out.println("Called unimplemented method: setSize: " + dim + ", " + size);
```
vfiack, line 178: is this setter necessary ?

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
```diff
@@ -17,987 +17,569 @@
...
-        return (fttype[iDim].equals("negate"));
+    public String getSequence() {
+        return block.optional(_PULSE_SEQUENCE, $PULPROG)
+                .map(JCampRecord::getString).orElse("");
     }
 
     @Override
-    public boolean getNegateImag(int iDim) {
-        return iDim > 0;
+    public double getSF(int dim) {
```
vfiack, line 202: Should SF be BASE_FREQ (without offset) or OBSERVED_FREQ? I chose the latter when possible.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
```diff
@@ -17,987 +17,569 @@
...
     @Override
-    public double getSW(int iDim) {
-        return Sw[iDim];
+    public String[] getSWNames() {
+        //XXX check whether this is useful
+        // original JCAMPData returned an array of empty strings
+        String[] names = new String[getNDim()];
+        Arrays.fill(names, "");
+        names[0] = block.optional($SW_H).map(JCampRecord::getNormalizedLabel).orElse("");
+        return names;
```
vfiack, line 267: This was not implemented in the original. I can see the change in the UI, I'm not sure how useful it is.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
```diff
@@ -17,987 +17,569 @@
...
 
     @Override
-    public double getRefPoint(int dim) {
-        return 1.0;
-    }
+    public String getFTType(int dim) {
+        // known values: ft, rft (real), negate (hypercomplex)
+        //XXX original JCamp has "ft" hardcoded.
+        // Bruker is using AQ_Mod and FnMode, which is what I chose to copy here.
+        // Not whether it should be filled for FID as well.
```
vfiack, line 336: Not whether it should be filled for FID as well. Would be better with an enum instead of a string, but this is out of scope here.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
```diff
@@ -17,987 +17,569 @@
...
 
     @Override
-    public double getTempK() {
-        Double d;
-        if ((d = getParDouble("TEMPERATURE,1")) == null) {
-// fixme what should we return if not present, is it ever not present
-            d = 298.0;
+    public double[] getCoefs(int dim) {
+        //XXX Was not implemented in original JCAMPData.
+        //Inspired from BrukerData instead.
```
vfiack, line 357: Not completely sure what this is supposed to do, especially because it was not implemented in the original JCAMPData.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
```diff
@@ -17,987 +17,569 @@
...
-            try {
-                System.out.println("sec " + s);
-                seconds = Long.parseLong(s);
-            } catch (NumberFormatException e) {
-            }
-        } else {
-            System.out.println("no date");
+    public String getSymbolicCoefs(int dim) {
+        //XXX Was not implemented in original JCAMPData.
+        //Inspired from BrukerData instead.
```
vfiack, line 378: Same: Not completely sure what this is supposed to do, especially because it was not implemented in the original JCAMPData.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
```diff
@@ -17,987 +17,569 @@
...
-                ph0 += 90.0;
-            } else if (iDim == 1) {
-                ph0 += deltaPh0_2;
-            }
-        }
-        return ph0;
+    public String getVendor() {
+        //XXX some code expected "bruker" in lowercase
+        // RefManager.setupItems() to add getNDim() in a string builder
+        // RefManager.setupItems() to add a getFixDSP() option
```
vfiack, line 402: Returning something other that "bruker" should be correct but will have some behavior change

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
```diff
@@ -17,987 +17,569 @@
...
+                .orElse(0.0);
 
-    @Override
-    public int getLeftShift(int iDim) {
-        int shift = 0;
-        Integer ipar;
-        if ((ipar = getParInt("LS," + (iDim + 1))) != null) {
-            shift = -ipar;
+        if (dim == 0) {
+            ph0 -= 90; //XXX from original JCAMPData, but I don't know why
```
vfiack, line 415: Why the 90° shift?

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
```diff
@@ -17,987 +17,569 @@
...
             }
         } else {
-            ArrayList<Double> iValues = iparser.getYValues();
+            double[] iValues = imaginaryPages.get(index).toArray();
             dvec.resize(n, true);
             dvec.setTDSize(n);
             for (int i = 0; i < n; i++) {
-                dvec.set(i, iValues.get(i), rValues.get(i));
+                //WARNING: real and imaginaries are inverted on purpose
+                dvec.set(i, iValues[i], rValues[i]);
```
vfiack, line 548: This is what inverted the spectra.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
```diff
@@ -17,987 +17,569 @@
...
-        }
     }
 
     @Override
     public void readVector(int iVec, Complex[] cdata) {
-        byte[] dataBuf = new byte[tbytes];
-        readVecBlock(iVec, dataBuf);
-        copyVecData(dataBuf, cdata);
+        //TODO implement me? needed?
+        throw new UnsupportedOperationException("Not yet implemented: readVector(int iVec, Complex[] cdata)");
```
vfiack, line 564: All these readVector() alternatives are never called, as far as I could see while testing. If they are only be called by other NMRData implementation internally, we should remove them from the interface. Otherwise, I can implement them if it is useful.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
```diff
@@ -1007,11 +589,12 @@ public void resetAcqOrder() {
 
     @Override
     public String[] getAcqOrder() {
+        //XXX I have no idea what this tries to do.
```
vfiack, line 592: Unclear what it is, but has an impact on MultiVecCounter, vector groups, and ultimately vector indexes, which currenly causes me problem when processing 2D files.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
```diff
@@ -1056,58 +678,79 @@ public void setSampleSchedule(SampleSchedule sampleSchedule) {
...
+    public boolean getNegatePairs(int dim) {
+        //XXX doesn't seem to be called. Is this useful?
+        return "negate".equals(getFTType(dim));
     }
 
-    class BrukerGaussianWt extends GaussianWt {
+    @Override
+    public boolean getNegateImag(int dim) {
+        //XXX doesn't seem to be called. Is this useful?
+        return dim > 0;
```
vfiack, line 700: Is this called somewhere?

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
```diff
@@ -1056,58 +678,79 @@ public void setSampleSchedule(SampleSchedule sampleSchedule) {
...
-    public static void main(String[] args) {
+    @Override
+    public String toString() {
+        return getFilePath();
+    }
 
+    private FnMode getFnMode(int dim) {
+        int value = block.optional($FN_MODE, dim).map(JCampRecord::getInt).orElse(-1);
+        if(value < 0 || value >= FnMode.values().length)
+            return null; // XXX should really be "no FnMode defined" be different from FnMode == "0/undefined"?
```
vfiack, line 721: The original and bruker parsers have different behavior for FnMODE not defined and FnMODE = 0 (undefined).

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
brucejohnson: getNVectors is the total number of vectors of data in dataset.  1 for 1D, nRows for 2D, nRows*nPlanes for 3D etc.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
brucejohnson: The coefs are used in combining data rows for effects like echo-antiecho.  The concept of coefs is originally inspired by Varian processing which used coefs used to create a new vector from a linear combination of original vectors in data.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
brucejohnson: Different vendors collect data in different orders (which is especially important for 3D and higher).  If d refers to t1, t2, etc delays, and p refers to phases (real imaginary) and the number refers to the dimension.
Bruker might collect p1,d1,p2,d2
Varian might collect p1,p2,d1,d2
where the first item is incremeted fastest.
This symbolic pattern is used to setup the MultiVecCounter which determines the position vectors are read from dataset.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
brucejohnson: Yes, that was a bug, though it looks like getNPoints wasn't actually used here.  The size was determined in readVEctor from the number of yValues parsed

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
brucejohnson: This is used in setting up processing scripts so that rather than setting an explcit value for the sweep width (or spectrometer frequency) one can specify the name of the parameter to use.  That way the same script can be used with different conditions.  Some pulse programs (I remember doing this on Varian) used different parameters for a given dimension.  Some times the first indirect dimension might have used "sw1" and other times "sw".  Probably not necessary to support here.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
brucejohnson: This is used in setting up processing scripts in python code so while an enum would be better in Java, the string was certainly simpler in Python (probably a way to use enums in Jython, but I"m not sure I've done it).

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
brucejohnson: I should probably review all these variants of readVector.  They go way back to some original versions where there were different choices for different data file arrangements etc.  We can probably get rid of most of them, but I better look pretty carefully at this.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/JCAMPData.java
```diff
@@ -1056,58 +678,79 @@ public void setSampleSchedule(SampleSchedule sampleSchedule) {
...
-            }
-        }
+    @Override
+    public void setPreferredDatasetType(DatasetType datasetType) {
+        this.preferredDatasetType = datasetType;
+    }
 
+    @Override
+    public boolean getNegatePairs(int dim) {
+        //XXX doesn't seem to be called. Is this useful?
```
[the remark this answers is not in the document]
brucejohnson: In Python code when auto-generating a script.

brucejohnson (approved):

PR #74, opened 2022-03-09
Infer SW from FIRST/LAST timestamps when no SW record in jcamp file

vfiack:


brucejohnson (approved):
I made one change in the python test script for the camp file to remove wrong names from the sw() and sf() commands.  Now the test passes

PR #91, opened 2022-04-04
NMR-4963: publish builds to nanalysis's sharepoint

vfiack:
Ready for review.

### DIFF
--- a/.github/workflows/actions-publish-build.sh
+++ b/.github/workflows/actions-publish-build.sh
@@ -1,0 +1,34 @@
+#!/bin/sh
+
+SHAREPOINT="java -jar sharepoint-client-1.3-jar-with-dependencies.jar"
+HOST="https://nanalysis.sharepoint.com"
+SITE="SoftwareDev"
+PARENT_PATH="Shared Documents/Builds/nmrfx"
+# client id and secret are passed through environment
+
+TODAY=`date +%Y%m%d`
+
+# get branch name from PR head if this is a PR workflow
+BRANCH=${GITHUB_HEAD_REF##*/}
+if [ -z "$BRANCH" ]; then
+  # otherwise, get the ref name
+  BRANCH=${GITHUB_REF##*/}
+fi
+
+# exits on error
+set -e
+
+# prepare local folder, in a temp folder managed by github action to ensure it's removed at the end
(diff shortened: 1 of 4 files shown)


brucejohnson:
If I understand correctly this will push to a private Nanalysis Sharepoint site.  Is it OK to be in the public repository?

vfiack:
> If I understand correctly this will push to a private Nanalysis Sharepoint site. Is it OK to be in the public repository?

You understand correctly. The credentials are stored as a repository secret and are not accessible, even on a public repository, so I think it is ok. The only issue I can think of is if we grant an external user write access to this repository, and they decide to create lots of pull requests to waste storage space.

.github/workflows/actions-publish-build.sh
```diff
@@ -0,0 +1,34 @@
...
+LOCAL_FOLDER_NAME=`echo "archive-${TODAY}-${BRANCH}-${GITHUB_RUN_NUMBER}" | tr "/" "_"`
+LOCAL_FOLDER="${RUNNER_TEMP}/${LOCAL_FOLDER_NAME}"
+mkdir "${LOCAL_FOLDER}"
+cp **/target/*.zip "${LOCAL_FOLDER}"
+cp nmrfx-jmx-connector/target/*.jar "${LOCAL_FOLDER}"
+
+# upload to remote, deleting first
+REMOTE_FOLDER=`echo "${BRANCH}" | tr "/" "_"`
+$SHAREPOINT "${HOST}" "${SITE}" api "${SHAREPOINT_CLIENT_ID}" "${SHAREPOINT_CLIENT_SECRET}" delete-folder "${PARENT_PATH}/${REMOTE_FOLDER}"
+$SHAREPOINT "${HOST}" "${SITE}" api "${SHAREPOINT_CLIENT_ID}" "${SHAREPOINT_CLIENT_SECRET}" upload-folder "${LOCAL_FOLDER}" "${PARENT_PATH}" "${REMOTE_FOLDER}"
```
darbymcgraw, line 31: How come this folder is deleted?

vfiack: Not sure what the question is:
* We delete it before upload to avoid having a mix of two different builds on the same branch name when you push again on an open PR or merge master.
* The folder is finally deleted by "cleanup" workflow on branch deletion.

brucejohnson (approved):

PR #94, opened 2022-04-08
Fix slices and vector extract in spinit data

brucejohnson:


vfiack (approved):
Working as expected.
Was the issue only on spinit data? The fix doesn't look too specific.

brucejohnson:
The issue was only happening on spirit data, but could in principle happen with other data.  The problem was both due to a reversed axis and also the complex data in first dimension.  There was also a problem with extracting slices to display in a second window.  Some day I'd like to rethink how complex, processed data is handled, but this works for now.

PR #105, opened 2022-05-03
Nmr 5105 sonar identical branches

JannalieT:


nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/PropertyManager.java
```diff
@@ -753,13 +753,7 @@ void setupItems() {
...
                             //ListOperationItemTypeSelector typeSelector = new ListOperationItemTypeSelector(stringListener, (String) listTypes.get(0), listTypes, op, "listType", parDesc);
                             ListOperationItemTypeSelector typeSelector = null;
-
-                            if (listTypes == null || listTypes.isEmpty()) { // just a default list, so treat them as real values.
-                                lstItem = new ListOperationItem(listListener, defaultList, listTypes, op, name, parDesc, typeSelector);
-                            } else {
-                                lstItem = new ListOperationItem(listListener, defaultList, listTypes, op, name, parDesc, typeSelector);
-                            }
-
+                            lstItem = new ListOperationItem(listListener, defaultList, listTypes, op, name, parDesc, typeSelector);
```
JannalieT, line 756: The ListOperationItem was created the same way regardless of the value of list items, so I collapsed the if statement. In ListOperationItem constructor, listTypes is assigned to ListOperationItem.listTypes which doesn’t seem like it is used anywhere (
https://github.com/nanalysis/nmrfx/blob/7d9b8e15b120fd9ddb133501a913dcd9801e67fe/nmrfx-utils/src/main/java/org/nmrfx/utils/properties/ListOperationItem.java#L67, 
https://github.com/nanalysis/nmrfx/blob/7d9b8e15b120fd9ddb133501a913dcd9801e67fe/nmrfx-utils/src/main/java/org/nmrfx/utils/properties/ListOperationItem.java#L45
)
should this variable be removed from ListOperationItem?

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/RefManager.java
```diff
@@ -246,11 +246,7 @@ String getPropValue(int dim, String propName, boolean getDefault) {
                     }
                     break;
                 case "acqarray":
-                    if (getDefault) {
-                        value = "0";
-                    } else {
-                        value = "0";
-                    }
+                    value = "0";
```
JannalieT, line 249: I am unsure on this one whether I should have removed of fixed the if/else. In relation to this on Line 360 (https://github.com/nanalysis/nmrfx/blob/7d9b8e15b120fd9ddb133501a913dcd9801e67fe/nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/RefManager.java#L360), the default value comment is hardcoded to 0, should the default value be retrieved by the line `String defaultValue = getPropValue(dim, probName, true);` ? It also seems that getPropValue is never really called for “acqarray” since the other place it might have been called avoids calling it (https://github.com/nanalysis/nmrfx/blob/7d9b8e15b120fd9ddb133501a913dcd9801e67fe/nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/RefManager.java#L172)

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/peaks/PeakListTools.java
```diff
@@ -1497,13 +1497,8 @@ public static List<Object> peakFit(PeakList peakList, Dataset theFile, List<Peak
...
         }
-        if ((delays != null) && (delays.length > 0)) {
-            GuessValue gValue = new GuessValue(0.0, -0.5 * globalMax, 0.5 * globalMax, false);
-            guessList.add(0, gValue);
-        } else {
-            GuessValue gValue = new GuessValue(0.0, -0.5 * globalMax, 0.5 * globalMax, false);
-            guessList.add(0, gValue);
-        }
+        GuessValue gValue = new GuessValue(0.0, -0.5 * globalMax, 0.5 * globalMax, false);
+        guessList.add(0, gValue);
```
JannalieT, line 1501: removing the if/else resulted in the variable needed to be defined in the scope, so changed a few more lines below this to fix redeclaring the variable

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/bruker/BrukerData.java
```diff
@@ -857,19 +853,7 @@ private void openParFile(File parDirFile) throws IOException {
...
-                                if (iPar > 1) {
-                                    acqdim++;
-                                } else {
-                                    acqdim++;
-                                }
-                            } else {
-                                acqdim++;
-                            }
-                        }
+                        acqdim++;
```
JannalieT, line 856: Removing the identical branch resulted in another identical branch which resulted in another identical branch, so removed all three since acqdim always increased

nmrfx-processor/src/main/java/org/nmrfx/processor/processing/SampleSchedule.java
```diff
@@ -402,12 +402,7 @@ public VecIndex getNextGroup(final int vecGroup, MultiVecCounter tmult) {
         }
         if (k < nSamples) {
             int[] oSizes;
-            if (demo) {
-                oSizes = tmult.getOutSizes();
-            } else {
-//                oSizes = outMult.getOutSizes();
-                oSizes = tmult.getOutSizes();
-            }
+            oSizes = tmult.getOutSizes();
```
JannalieT, line 405: Collapsed the if loop but not sure if this is correct, looks like outMult.getOutSizes was commented out to just use tMult, if this change is correct the function comment needs to be updated as it contradicts what is happening (https://github.com/nanalysis/nmrfx/blob/7d9b8e15b120fd9ddb133501a913dcd9801e67fe/nmrfx-processor/src/main/java/org/nmrfx/processor/processing/SampleSchedule.java#L383)

Maybe remove everything in the comment after the first sentence?

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/predict/HosePrediction.java
```diff
@@ -609,18 +609,11 @@ void validate(String validate) {
...
                         System.out.printf("%s\t%8.2f\t%8.2f\t%8.2f\t%d\t%d\t%s\t%s\t%s\n", statNames[i], ppm, predPPM, Math.abs(ppm - predPPM), iShell, stat.nValues, hoseString[0], smile, molNum);
                     }
-                    if ((stat == null) || (stat.nValues < 4)) {
-                        if (stat == null) {
-                            //System.err.println(lineReader.getLineNumber() + " " + statNames[i] + " " + hoseString[0] + " null " + iShell + " " +  ppm);
-                        } else {
-                            //System.err.println(lineReader.getLineNumber() + " " + statNames[i] + " " + hoseString[0] + " insufficient " + stat.nValues + " " +  iShell + " " + ppm);
-                        }
-                    } else {
+                    if (!((stat == null) || (stat.nValues < 4))) {
```
JannalieT, line 612: removed commented out logs in both the if and else, and converted the else to if

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/bruker/BrukerData.java
```diff
@@ -112,11 +112,7 @@ public BrukerData(String path, File nusFile) throws IOException {
         File file = new File(path);
         openParFile(file);
         openDataFile(path);
-        if (dim < 2) {
-            scale = 1.0e6;
-        } else {
-            scale = 1.0e6;
-        }
+        scale = 1.0e6;
```
JannalieT, line 115: removed the if/else since seems to be in the file from the beginning but double check if scale is actually supposed to be different?

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/peaks/PeakListTools.java
vfiack: I'd like to avoid reusing variables if we can. You renamed to `gVal` line 1549 which is good, could you do something similar line 1529, or inline the variable here?

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/bruker/BrukerData.java
vfiack: This constructor is called only for FID as far as I can see. The other constructor sets scale to `1` and is called for processed data. @brucejohnson can you confirm?

nmrfx-processor/src/main/java/org/nmrfx/processor/processing/SampleSchedule.java
vfiack: It looks like this method isn't even used. I'd remove it (as well as other unused methods and commented code, in general). We need to be careful because some methods can be used from python script and are not detected by tools such as intellij, but I think this one is safe?

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/predict/HosePrediction.java
vfiack, line 612: `x != y` is easier to read that `!(x == y)` 😄  

It also looks like you inverted the logic here.

```suggestion
                    if ((stat != null && stat.nValues >= 4)) {
```

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/PropertyManager.java
vfiack: I would remove it - the less code we keep, the less bugs we risk. If it was useful, someone would have noticed that it didn't do anything since at least 2018.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/peaks/PeakListTools.java
JannalieT: used inline variables and renamed another reuse of gValue as well

nmrfx-processor/src/main/java/org/nmrfx/processor/processing/SampleSchedule.java
JannalieT: Removed this method

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/predict/HosePrediction.java
JannalieT: Oops! Changed to your suggestion.

brucejohnson (dismissed):

nmrfx-utils/src/main/java/org/nmrfx/utils/properties/ListOperationItem.java
```diff
@@ -35,38 +37,31 @@
...
      * @param defaultValue optional default value for the List.
-     * @param listTypes The types of objects that will be contained in the list.
-     * The List can only support one type of item at a time.
      * @param category
      * @param name
      * @param description
      * @param typeSelector
      */
-    public ListOperationItem(ChangeListener listener, ArrayList defaultValue, ArrayList<String> listTypes, String category, String name, String description, ChoiceOperationItem typeSelector) {
+    public ListOperationItem(ChangeListener<? super String> listener, List<String> defaultValue, String category, String name, String description, ChoiceOperationItem typeSelector) {
```
JannalieT, line 56: Made these changes because sonar wanted a type specified and to use List over ArrayList

vfiack: `? super String` is strange, because String is final (meaning, there can be no subclasses).

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/bruker/BrukerData.java
brucejohnson: This change can be left in, but after this PR is merged I'm going to do some work on setting scale.  There is a Bruker parameter that indicates how the data was scaled during collection and we are ignoring that parameter.  That results in relative intensities being wrong when multiple datasets are collected on same sample (recently pointed out by user).

nmrfx-utils/src/main/java/org/nmrfx/utils/properties/ListOperationItem.java
JannalieT: Yeah I wasn't sure about this one, sonar wanted a type so I used the type that was used a few lines above: https://github.com/nanalysis/nmrfx/blob/dd83badf98ccf0847d0c0d02a7c58661a78b1228/nmrfx-utils/src/main/java/org/nmrfx/utils/properties/ListOperationItem.java#L39
I think this one might actually be String and its super Object, according to this https://docs.oracle.com/javase/tutorial/java/generics/lowerBounded.html

brucejohnson (approved):

nmrfx-utils/src/main/java/org/nmrfx/utils/properties/ListOperationItem.java
```diff
@@ -53,10 +53,10 @@ public class ListOperationItem extends OperationItem implements ObservableObject
      * @param description
      * @param typeSelector
      */
-    public ListOperationItem(ChangeListener<? super String> listener, List<String> defaultValue, String category, String name, String description, ChoiceOperationItem typeSelector) {
+    public ListOperationItem(ChangeListener<? super String> listener, List<?> defaultValue, String category, String name, String description, ChoiceOperationItem typeSelector) {
         super(category, name, description);
         if (defaultValue != null) {
-            this.defaultValue = (ArrayList<String>) defaultValue;
+            this.defaultValue = (ArrayList<?>) defaultValue;
```
vfiack, line 59: This is dangerous as well: someone could create a ListOperationItem with another List implementation, and that would crash with a ClassCastException. 

Several possibilties:
* change the field to List<?> if we can
* check the type with instanceof before cast
* ensure it's an arraylist by creating a new one before assignment

Of those, the best one would be to change the types from ArrayList to List but there are some impacts (getters should change as well, and their usage must be checked) and this is getting a bit far from the initial goal of this PR.

vfiack (approved):

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/PropertyManager.java
```diff
@@ -339,37 +339,6 @@ private void updateOp(PropertySheet.Item updateItem) {
 
     }
 
-    void addRegionRangeOld(double min, double max, double f1, double f2) {
-        System.out.println("add region range " + f1 + " " + f2);
```
JannalieT, line 343: Unused so removed

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/PropertyManager.java
```diff
@@ -748,23 +717,11 @@ void setupItems() {
 
                             break;
                         case "list":
-                            ArrayList listTypes = (ArrayList) parMap.get("listTypes");
                             ListOperationItem lstItem;
                             ArrayList defaultList = (ArrayList) parMap.get("default");
```
JannalieT, line 721: Did not do anything with this line, not sure if I should have?

nmrfx-utils/src/main/java/org/nmrfx/utils/properties/ListOperationItem.java
```diff
@@ -55,25 +56,20 @@ public class ListOperationItem extends OperationItem implements ObservableObject
...
 
+    /**
+     * Returns values as a string created by the java List implementation of toString()
+     */
     @Override
     public String getValue() {
         return value.toString();
     }
 
-    public ArrayList getValueList() {
```
JannalieT, line 78: Only used by an unused function and was not an overriden method so removed

nmrfx-utils/src/test/java/org/nmrfx/utils/properties/ListOperationItemTest.java
```diff
@@ -0,0 +1,170 @@
...
+        changeListenerCallCount = 0;
+        listListener = new ChangeListener<String>() {
+            @Override
+            public void changed(ObservableValue<? extends String> observableValue, String c1, String c2) {
+                changeListenerCallCount++;
+            }
+        };
+
+    }
+    @Test
```
JannalieT, line 30:
```suggestion
    }
    
    @Test
```

nmrfx-utils/src/main/java/org/nmrfx/utils/properties/ListOperationItem.java
```diff
@@ -120,44 +110,50 @@ public void setValue(Object o) {
...
                     }
                 }
             }
+            newValue = numberValues;
         } else if (o instanceof ArrayList) {
-            newValue = new ArrayList((ArrayList) o);
+            newValue = new ArrayList<>((ArrayList<?>) o);
         } else {
             return;
         }
```
JannalieT, line 133: Value was never set unless a string or an array list was passed in, left this functionality as is.

nmrfx-utils/src/main/java/org/nmrfx/utils/properties/ListOperationItem.java
JannalieT: Changed the value and defaultValue fields to List<?>

nmrfx-utils/src/test/java/org/nmrfx/utils/properties/ListOperationItemTest.java
```diff
@@ -0,0 +1,171 @@
...
+    public void testSetValueArrayListSameValues() {
+        ArrayList<Double> defaultList = new ArrayList<>(List.of(1.0, 2.0, 3.0, 4.0));
+        ListOperationItem lstItem = new ListOperationItem(listListener, defaultList, "COADD", "coef", "List of coefficients to scale each vector by.", typeSelector);
+        ArrayList<Double> newValue = new ArrayList<>(List.of(1.0, 2.0, 3.0, 4.0));
+        lstItem.setValue(newValue);
+        Assert.assertEquals(List.of(1.0, 2.0, 3.0, 4.0), lstItem.value);
+        Assert.assertEquals(lstItem.value, lstItem.defaultValue);
+        // Listener is not updated if the values have not changed
+        Assert.assertEquals(0, changeListenerCallCount);
```
vfiack, line 117: Good initiative adding unit tests!

Minor things: 
* you could `import static import static org.junit.Assert.*` to avoid prefixing all assertions
* you can replace your comment by a message in the assertion (as the first parameter to asserEquals)

vfiack (approved):

nmrfx-utils/src/main/java/org/nmrfx/utils/properties/ListOperationItem.java
vfiack: Which is fine for this PR. 

In general though, I think this pattern is quite bad, and it should throw an `IllegalArgumentException` instead. I'm not sure whether this will break anything with existing code, so letting it as-is for now is ok. 

Maybe changing the instanceof check to accept any `List` type instead of only `ArrayList`, since we create a new one anyway, but that's again increasing this PR scope.

nmrfx-utils/src/test/java/org/nmrfx/utils/properties/ListOperationItemTest.java
JannalieT: Made these two changes

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/PropertyManager.java
JannalieT: Okay, I will leave it as is for this pr.

vfiack (approved):

PR #106, opened 2022-05-04
Nmr 5099 evaluate fxribbon

vfiack:
Prototype only, not to be merged in this state.
Still missing "molecule" menu.

Icons from "famfamfam silk" and "fatcow" icon sets.


vfiack:
Will not be done, we have another plan for UI simplification.

PR #107, opened 2022-05-04
Nmr 5106 sonar identical expressions

JannalieT:


nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/peaks/PeakNeighbors.java
```diff
@@ -184,46 +184,6 @@ final void setCells(int iP) {
...
-                    int ip = peakIndex[0][i];
-                    for (int j = jStart; j < jEnd; j++) {
-                        int jp = peakIndex[1][j];
-                        if ((iCell == jCell) && (ip >= jp)) {
-                            continue;
-                        }
-                        Peak peak1 = listPeaks.get(ip);
-                        Peak peak2 = listPeaks.get(jp);
-                        if ((peak1.getStatus() >= 0) && (peak1.getStatus() >= 0)) {
-                            double distance = peak1.distance(peak2, cellSizes);
```
JannalieT, line 219: when fixing this typo, seems like this method does not do anything and its also not called so removed the whole method

nmrfx-processor/src/main/java/org/nmrfx/processor/math/Polynomial.java
```diff
@@ -726,59 +720,12 @@ public void svejgardRoot(int nn, Complex[] poly) {
...
     }
 
-    private static void printCVec(String name, double[] ar, double[] ai) {
-        double avg = 0.0;
-        double sig = 0.0;
-        double max = 0.0;
-        double min = Double.MAX_VALUE;
-        Complex[] cValues = new Complex[ar.length];
-        if (0 == 0) {
-            return;
```
JannalieT, line 747: this method was not being used so removed it

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/miner/NanoMol.java
```diff
@@ -739,435 +665,6 @@ byte[] getBytes(byte[] inputBytes, int start) {
...
-
-    public void unCompressBondProps() throws Exception {
-        int j = 0;
-        BitSet bondBits = fromBytesToBitSet(bondBytes);
-        int nBondBits = getNBits(bondBytes);
-
-        for (int i = 0; i < nBondBits; i++) {
-            if (!bondBits.get(i)) {
-                nanoBonds[j++].bondProp = 1;
-            } else if (!bondBits.get(++i) && !bondBits.get(++i)) {
```
JannalieT, line 1045: -	Caught by sonar but I think these lines are technically not identical because the index has increased so they are retrieving a different bit, I think this function is a bit sketchy as well because of the way its incrementing i within the loop so there could be indexing errors, since this function is only called by one function, uncompressNanoRep which is not called at all, therefore I decided to remove this function and all other functions related to uncompress and compressing NanoRep, I put this in a separate commit in case removing all these methods is not wanted. I actually think the entire file could be removed as it does not seem to be used. The only other reference to NanoMol looks like it is in TurboMol.java (https://github.com/nanalysis/nmrfx/blob/7d9b8e15b120fd9ddb133501a913dcd9801e67fe/nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/miner/TurboMol.java#L293) and those methods do not appear to be implemented (https://github.com/nanalysis/nmrfx/blob/7d9b8e15b120fd9ddb133501a913dcd9801e67fe/nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/miner/TurboMol.java#L196) …Would it be okay to remove the whole file?

brucejohnson: I think we should actually leave NanoMol for now, and not even delete full unused methods.  We currently embed NMR information on a library of chemical compounds in NMRFx.  It would be good to have for each compound a drawing of the structure and the NanoMol code was designed to store that information in a highly compressed format.  So there is a reasonable chance that I will get around to using this code in the next few months.  If we decide to not do that, or do it in a different way, then we should delete it.

brucejohnson: I think we should actually leave NanoMol for now, and not even delete full unused methods.  We currently embed NMR information on a library of chemical compounds in NMRFx.  It would be good to have for each compound a drawing of the structure and the NanoMol code was designed to store that information in a highly compressed format.  So there is a reasonable chance that I will get around to using this code in the next few months.  If we decide to not do that, or do it in a different way, then we should delete it.

JannalieT: @vfiack If I leave this function in, should I just add a note that the sonar identical expressions error is not actually identical, and consider it 'fixed'?

vfiack: You can mark it as false positive on sonar directly, this will update it on github as well.

darbymcgraw (approved):

vfiack (approved):

PR #110, opened 2022-05-05
Nmr 5127 scanner tool improvements for pseudo 3 d

brucejohnson:


nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/peaks/PeakListTools.java
```diff
@@ -1056,10 +1056,13 @@ private static void setValues(Peak peak, double[][] values, String mode) {
     private static void measurePlanes(int nPlanes, Peak peak, Dataset dataset,
             java.util.function.Function<RegionData, Double> f,
             String mode, double[][] values, int iValue) {
-        int[] planes = new int[1];
+        int extraPlanes = dataset.getNDim() - peak.getNDim();
+        int[] planes = new int[extraPlanes];
```
JannalieT, line 1060: Just wanted to check that extraPlanes would never be negative?

brucejohnson: I don't think that would ever happen, but I've added a check to make sure extraPlanes is >=0 in case it does

JannalieT (approved):

PR #111, opened 2022-05-05
Get python tests working from junit

brucejohnson:


nmrfx-processor/src/test/python/testoperations.py
```diff
@@ -975,5 +975,10 @@ def testZFInverse(self):
         self.assert_(size == size3)
 
 
+#if __name__ == '__main__':
+#    unittest.main()
+
```
JannalieT, line 980: I think the commented out lines should be deleted if they aren't needed anymore
```suggestion
```

nmrfx-processor/src/test/python/testvec.py
```diff
@@ -384,7 +384,7 @@ def testResize(self):
         #FIXME test vector.pt on resize of real / complex Vecs
         import warnings
         import org.nmrfx.math.IllegalVecState as IllegalVecState
-        warnings.warn("testResize doesn't test vector.pt on resize")
+        #warnings.warn("testResize doesn't test vector.pt on resize")
```
JannalieT, line 387: Delete if not needed I think
```suggestion
```

nmrfx-structure/src/test/python/runtests.py
```diff
@@ -1,15 +0,0 @@
...
-import java.lang.RuntimeException
-import sys
-
-#all test files have to be loaded into the current namespace to be able to
-# execute them with unittest.main(), so we have to do: "from x import *"
-#jython script
-from teststruct import *
-
-if __name__ == "__main__":
-    unittest.main()
```
JannalieT, line 11: I am a bit confused by the naming of the files, in the processor test python directory there is a file runtests.py, but in this structure test python directory the runtests.py is deleted. Does runtests.py just run all the test scripts in the directory? and this is not needed here because there is only one python test script in structures?

nmrfx-processor/src/test/python/testvec.py
brucejohnson: Removed

nmrfx-processor/src/test/python/testoperations.py
brucejohnson: Lines deleted

nmrfx-structure/src/test/python/runtests.py
brucejohnson: These were leftover from old way of running python tests.  I've now removed the runtests.py script (and a couple others that should have been removed)

JannalieT (approved):

PR #115, opened 2022-05-09
Add release doc

vfiack:
@brucejohnson please tell me whether this is clear enough or if steps are missing or should be more detailed.


brucejohnson:
We'll see if we need to refine the protocol and description as we move forward, but it looks good to me.

PR #116, opened 2022-05-09
Change to version 11.2.4-a for release

brucejohnson:


vfiack:
Not to be merged in this state: the pom.xml files would move master from 11.2.5-SNAPSHOT to 11.2.4-a

vfiack (approved):
I reset the pom before merge. That way master stays in 11.2.5-SNAPSHOT, we can merge and remove the release branch, and the tag will remain in history.

PR #119, opened 2022-05-12
NMR-5154 fix reading/writing of deuterium relaxation parameters

brucejohnson:


nmrfx-core/src/main/java/org/nmrfx/chemistry/io/NMRStarReader.java
```diff
@@ -1862,18 +1862,14 @@ public void process(String[] argv) throws ParseException, IllegalArgumentExcepti
...
-                System.err.println("process T1rho");
-            }
-            buildRelaxation(relaxTypes.T1RHO);
-            if (DEBUG) {
-                System.err.println("process T2");
+            for (var relaxType: relaxTypes.values()) {
+                if ((relaxType != relaxTypes.NOE) && (relaxType != relaxTypes.S2)) {
+                    if (DEBUG) {
+                        System.err.println("process " + relaxType);
+                    }
```
vfiack, line 1869: This could be replaced with a debug log.

brucejohnson: I've replaced this (and all similar) statements with log.debug().  Also did some general code cleanup (mostly removing redundant  casts)

JannalieT (approved):

PR #121, opened 2022-05-18
Nmr 5102 avoid npe

JannalieT:


nmrfx-analyst/src/main/java/org/nmrfx/analyst/compounds/CompoundFitter.java
```diff
@@ -1045,17 +1049,21 @@ public ArrayList<FitResult> optimizeAlignment() {
...
-                cR.cMatch.setShift(cR.regions[iRegion], fitResult.getShift());
-                cR.shifts[iRegion] = fitResult.getShift();
+                // if fitResult is null don't add it to best shifts
+                if (fitResult != null) {
+                    bestShifts.add(fitResult);
+                    cR.cMatch.setShift(cR.regions[iRegion], fitResult.getShift());
+                    cR.shifts[iRegion] = fitResult.getShift();
+                } else {
+                    log.warn("Unable to find a best fit while optimizing alignment.");
+                }
```
JannalieT, line 1059: Wasn't sure if I should throw an exception here or if it is okay to just not add a null fit result...I guess the issue is that the cR.regions and cR.shifts could have array elements that were not set in this case.... This method is used by the spectrum fitter "Optimize" button, if I throw an exception here instead, would we need to have a pop up saying that the optimization failed?

image

nmrfx-analyst/src/main/java/org/nmrfx/analyst/peaks/Multiplets.java
```diff
@@ -184,7 +184,6 @@ public static Optional<Multiplet> mergePeaks(List<Peak> peaks) {
                 }
             }
             firstPeakDim.getMultiplet().updateCoupling(comps);
```
JannalieT, line 186: I think this is a false positive by sonar, the check on line 168 ensures that there will always be atleast one iteration of the loop on line 170, which either add a new peakDim to peakDims or throw an uncaught exception, therefore there will always be at least one entry in peakDims when it does the loop on line 179 that will set firstPeakDim on the first iteration of the loop.

nmrfx-analyst/src/main/java/org/nmrfx/analyst/peaks/Multiplets.java
```diff
@@ -240,32 +239,27 @@ public static Optional<Multiplet> transferPeaks(Multiplet multiplet, List<Peak>
     }
 
     public static void addPeaksToMultiplet(Multiplet multiplet, double... ppms) {
-        int iDim = 0;
         PeakDim peakDim = getMultipletRoot(multiplet);
```
JannalieT, line 242: The only way peakDim will be null is if the method parameter multiplet is null. Threw an IllegalArgumentException here, confirm if this is okay? or better to just log a message here.

nmrfx-analyst/src/main/java/org/nmrfx/analyst/peaks/PeakFitting.java
```diff
@@ -158,19 +163,22 @@ public double jfitLinkedPeak(Peak peak, String fitMode) {
...
+                limits[0][0] = bounds[0];
+                limits[0][1] = bounds[1];
+                List<Peak> peaks = peak.peakList.locatePeaks(limits, dims);
+                List<PeakDim> peakDims = new ArrayList<>();
+                for (Peak peakA : peaks) {
+                    peakDims.add(peakA.getPeakDim(0));
+                }
+                value = fitPeakDims(peakDims, mode, bounds, fitMode);
+            } else {
+                log.warn("Unable to get bounds to fit peaks.");
```
JannalieT, line 178: Since the previous behaviour was to just print the exception, just added a check for bounds being null and if it was, logged a message

nmrfx-core/src/main/java/org/nmrfx/chemistry/MoleculeBase.java
```diff
@@ -1084,17 +1084,17 @@ public void writeXYZToPDB(String fileName, int whichStruct) throws IOException {
                         lastAtom = atom;
                     }
                 }
-                out.print(lastAtom.spatialSet.toTERString(i + 1) + "\n");
-
-                bondList.forEach((bAtom) -> {
+                if (lastAtom != null) {
+                    out.print(lastAtom.spatialSet.toTERString(i + 1) + "\n");
+                }
+                bondList.forEach(bAtom -> {
```
JannalieT, line 1090: lastAtom should only be null if there are no atoms in the MoleculeBase instance, or no atoms spatial sets were able to be converted to a string. Is it okay to leave this last line off of the file in that case?

nmrfx-core/src/main/java/org/nmrfx/chemistry/Polymer.java
```diff
@@ -373,24 +373,24 @@ public boolean isDNA() {
...
+                }
+                if (atom2 == null) {
+                    System.out.println("no atom2 " + cyclicClosers[i + 1]);
+                    atom2 = getFirstResidue().getAtom("H1");
+                }
+                if (atom1 != null && atom2 != null) {
+                    String distance = cyclicClosers[i + 2];
+                    String constraint = atom2.getFullName() + " " + atom1.getFullName() + " " + distance;
+                    constraints.add(constraint);
+                }
```
JannalieT, line 392: Is it fine to just not add the constraint to the list being returned or should I raise an exception here? This method is called only once here: https://github.com/nanalysis/nmrfx/blob/4145ec117ba89f6926ff2be1dc000cdcaaa25900/nmrfx-structure/src/main/resources/refine.py#L792

nmrfx-core/src/main/java/org/nmrfx/chemistry/io/PDBFile.java
```diff
@@ -218,6 +218,10 @@ public MoleculeBase read(String fileName)
                         molecule.addEntity(compound, coordSetName);
                     }
 
+                    if (compound == null) {
+                        throw new MoleculeIOException("didn't form compound");
+                    }
+
```
JannalieT, line 224: Followed same format as Polyer exception earlier in function: https://github.com/nanalysis/nmrfx/blob/4145ec117ba89f6926ff2be1dc000cdcaaa25900/nmrfx-core/src/main/java/org/nmrfx/chemistry/io/PDBFile.java#L175

nmrfx-core/src/main/java/org/nmrfx/chemistry/io/PDBFile.java
```diff
@@ -1041,15 +1045,15 @@ public static Compound readResidue(String fileName, String fileContent, Molecule
             } catch (IOException ioe) {
                 System.err.println(ioe.getMessage());
                 molecule.getAtomTypes();
-                if (calcBonds) {
+                if (calcBonds && compound != null) {
                     compound.calcAllBonds();
                 }
                 return compound;
```
JannalieT, line 1051: It seems a bit strange to return a compound that may not have had everything added to it from the file, it was specifically changed to do this in 2018 though so left it like this...

nmrfx-core/src/main/java/org/nmrfx/datasets/DatasetRegion.java
```diff
@@ -281,6 +281,10 @@ public int compare(Object o1, Object o2) {
                 result = -1;
             } else if (r2 == null) {
                 result = 1;
+            } else if (r1.x == null) {
+                result = -1;
+            } else if (r2.x == null) {
+                result = 1;
```
JannalieT, line 287: added these two conditions to avoid NPE. This change means that if two DatasetRegions have the parameter x as null, they will not be considered equal. Should they be considered equal? (before there would have been an exception thrown during the comparison). I didn't want to make too many changes as there is this note on the file indicating that the equals method may not be correct anyways: https://github.com/nanalysis/nmrfx/blob/4145ec117ba89f6926ff2be1dc000cdcaaa25900/nmrfx-core/src/main/java/org/nmrfx/datasets/DatasetRegion.java#L9

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/ChartProcessor.java
```diff
@@ -575,7 +575,7 @@ public Map<String, List<String>> getScriptList() {
         if (mapOpLists != null) {
             for (Map.Entry<String, List<String>> entry : mapOpLists.entrySet()) {
                 List<String> newList = new ArrayList<>();
-                if (entry != null) {
+                if (entry.getValue() != null) {
                     newList.addAll(entry.getValue());
                 }
```
JannalieT, line 580: I think in this case entry object itself should never be null?, but its value could be so its better to check that instead

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/FXMLController.java
```diff
@@ -603,30 +603,28 @@ public boolean isFIDActive() {
...
+            processorController.clearOperationList();
+            chartProcessor.clearAllOperations();
+            processorController.parseScript("");
+            if (!reload) {
+                getActiveChart().full();
+                getActiveChart().autoScale();
+                chartProcessor.loadDefaultScriptIfPresent();
+            }
+            getActiveChart().layoutPlotChildren();
+            statusBar.setMode(0);
```
JannalieT, line 624: moved possible NPE code into the null check and added a log message, not sure if I should do anything else here? the previous behaviour would have just not added the FID because of the NPE without any message to the user, should there be a dialog to the user?
Also no null check needed for chartProcessor because if ProcessorController.create creates the processorController then chartProcessor will not be null.

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/OperationListCell.java
```diff
@@ -160,26 +160,17 @@ public void handle(DragEvent event) {
...
-                    String swap;
-                    //ObservableList<String> listItems = MainApp.mainController.listItems;
-
-                    swap = listItems.remove(sourceIndex);
-                    listItems.add(targetIndex, swap);
-                    scriptView.getSelectionModel().select(targetIndex);
-// fixme
-//                    processorController.chartProcessor.execScriptList();
-//                    processorController.propertyManager.chartProcessor.getChart().layoutPlotChildren();
-//                    processorController.propertyManager.popOver.hide();
```
JannalieT, line 180: These do say fix me, but they have been commented out since 2017...I can add them back in if needed.

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/spectra/SpecRegion.java
```diff
@@ -142,6 +142,10 @@ public int compare(Object o1, Object o2) {
                 result = -1;
             } else if (r2 == null) {
                 result = 1;
+            } else if (r1.x == null) {
+                result = -1;
+            } else if (r2.x == null) {
+                result = 1;
```
JannalieT, line 148: see comment about DatasetRegion changes

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/varian/VNMRPar.java
```diff
@@ -115,10 +115,12 @@ public static LinkedHashMap<String, String> getParMap(String fpath) {
...
             String handle = processVNMRParFile(fpath);
             ArrayList<String> parNames = getPars(handle);
-            hmap = new LinkedHashMap();
+            hmap = new LinkedHashMap<>();
             for (String parName : parNames) {
                 VNMRPar vp = getPar(handle, parName);
-                hmap.put(parName, vp.getJValues());
+                if (vp != null) {
+                    hmap.put(parName, vp.getJValues());
+                }
```
JannalieT, line 123: Changed this to how possible NPE were handled in other overloaded version of function: https://github.com/nanalysis/nmrfx/blob/4145ec117ba89f6926ff2be1dc000cdcaaa25900/nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/varian/VNMRPar.java#L145

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/Molecule.java
```diff
@@ -804,38 +804,6 @@ public void updateFromVecCoords() {
...
-                    continue;
-                }
-                if (a4 == a3.parent) {
-                    continue;
-                }
-                if (a4.parent != a3) {
-                    continue;
-                }
-                if (dihedralAngles == null) {
-                    dihedralAngles[nAngles] = a4.dihedralAngle;
```
JannalieT, line 832: this was probably a typo for !=, but method was unused so removed method instead.

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/energy/CmaesRefinement.java
```diff
@@ -121,12 +121,12 @@ public double refineCMAES(final int nSteps, final double stopFitness, final doub
                     new InitialGuess(dihedrals.normValues));
         } catch (Exception e) {
             e.printStackTrace();
+            throw e;
```
JannalieT, line 124: asked @brucejohnson whether to rethrow these exceptions.

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/energy/Dihedral.java
```diff
@@ -429,41 +429,24 @@ public void setBoundaries(final double sigma, boolean useDegrees, double maxRang
...
-                    inputSigma[aStart++] = sigma / backBoneScale;
-                }
-            } else if (atom.parent.getName().equals("CA") || atom.parent.getName().equals("C")) {
-                inputSigma[aStart++] = sigma / backBoneScale;
-                if (sinCosMode) {
-                    inputSigma[aStart++] = sigma / backBoneScale;
-                }
+                throw new NullPointerException("daughter null " + angleAtoms.get(iAtom).daughterAtom.getFullName());
+            } else if (atom.parent == null) {
+                throw new NullPointerException("parent null " + angleAtoms.get(iAtom).daughterAtom.getFullName());
```
JannalieT, line 434: the if statements were specifically added as debug statements to be printed before the NPE occurred but nothing was done to stop the NPE from happening so seems like it is expected for them to be thrown here, moved the messages to the exceptions themselves. Not sure if this is okay?

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/energy/RNARotamer.java
```diff
@@ -417,23 +421,7 @@ public static void clear() {
     public static void add(String name, int n, double... angles) {
         ROTAMERS.put(name, new RNARotamer(name, n, angles));
     }
-
-    public static void validateDetail() {
-        for (RNARotamer rotamer : ROTAMERS.values()) {
```
JannalieT, line 422: Unused, and seem to just be for printing the scores

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/energy/RNARotamer.java
```diff
@@ -772,6 +760,10 @@ public static void setDihedrals(Residue residue, String suiteName) {
      * refinement
      */
     public static void setDihedrals(Residue residue, String suiteName, double sdev, boolean doFreeze) {
+        if (residue == null) {
+            log.warn("Residue is null. Unable to set Dihedrals.");
+            return;
+        }
```
JannalieT, line 766: Returned early from this method if residue was null, is this okay? Would it be better to throw and illegal argument exception here if residue is null? This method is only used in refine.py

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/energy/RotationalDynamics.java
```diff
@@ -200,12 +200,12 @@ public void advanceDihedrals2(double timestep) {
                 max = absDelta;
             }
             sumSq += delAngle * delAngle;
-            //System.out.println(deltaSum);
             if (daughter == null) {
-                System.out.println("daughter atom is null " + diAtom.getShortName());
+                throw new NullPointerException("daughter atom is null " + diAtom.getShortName());
```
JannalieT, line 204: This print statement was added specifically for debugging but nothing done about the NPE so assumed it is meant to be thrown here and moved the print statement to the exception message.

nmrfx-structure/src/main/java/org/nmrfx/structure/seqassign/SpinSystems.java
```diff
@@ -579,19 +579,19 @@ void readSTARSaveFrame(Saveframe saveframe) throws ParseException {
...
-                            nextSystem.confirmP = Optional.of(match);
+
+                    if (nextSystem != null) {
+                        // find match that is to next, confirm and add to a fragment
+                        for (SpinSystemMatch match : system.getMatchToNext()) {
+                            if ((match.getSpinSystemA() == system) && (match.getSpinSystemB() == nextSystem)) {
+                                fragment.getSpinSystemMatches().add(match);
+                                system.confirmS = Optional.of(match);
+                                nextSystem.confirmP = Optional.of(match);
+                            }
```
JannalieT, line 594: moved this within the fragment null check, confirm if this is okay?

JannalieT:
I think the NPE located in this method is a false positive https://github.com/nanalysis/nmrfx/blob/4145ec117ba89f6926ff2be1dc000cdcaaa25900/nmrfx-processor/src/main/java/org/nmrfx/processor/operations/IstMatrix.java#L210
orig is set to null on line 210, an if statement on line 211 with the condition `final_ift = true` sets orig and orig is not used until an if statement on line 227 with the same condition, not sure if this is okay  to mark as false positive, or should I replace line 210 with `orig = new Matrix(msize[0], msize[1]);` so that it will always be not null, even if final_ift is false?

JannalieT:
I think the NPE located in this method is a false positive

https://github.com/nanalysis/nmrfx/blob/4145ec117ba89f6926ff2be1dc000cdcaaa25900/nmrfx-processor/src/main/java/org/nmrfx/processor/math/IstMath.java#L143

orig is set to null on line 143, an if statement on line 147 with the condition `timeDomain = true` sets orig and orig is not used until an if statement on line 167 with the same condition, not sure if this is okay to mark as false positive, or should I replace line 143 with `orig = new Complex[len];` so that it will always be not null, even if timeDomain is false?

nmrfx-analyst/src/main/java/org/nmrfx/analyst/compounds/CompoundFitter.java
```diff
@@ -618,6 +620,7 @@ public double[] optimizeByCMAES(int start) {
                     new InitialGuess(starting[0]));
         } catch (DimensionMismatchException | NotPositiveException | NotStrictlyPositiveException | TooManyEvaluationsException e) {
             e.printStackTrace();
+            throw e;
```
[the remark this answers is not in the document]
vfiack: This looks more like code that isn't used yet (for 3 years) than obsolete code. @brucejohnson ?

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/ChartProcessor.java
```diff
@@ -685,11 +683,7 @@ public void setVecDim(String dimName) {
             if (mapOpLists.get(vecDimName) != null) {
                 oldList.addAll(mapOpLists.get(vecDimName));
             }
-            //execScriptList(false);
         } else {
-            if (mapOpLists == null) {
-
-            }
```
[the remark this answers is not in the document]
JannalieT: merged into else if

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/spectra/SpecRegion.java
JannalieT: merged tests

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/energy/Dihedral.java
JannalieT: Changed to use IllegalStateException

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/energy/GradientRefinement.java
```diff
@@ -132,7 +132,8 @@ public void gradMinimizeNLCG(int nSteps, double tolerance) {
                     GoalType.MINIMIZE,
                     new InitialGuess(dihedrals.angleValues));
         } catch (Exception e) {
-            e.printStackTrace();
+            log.error(e.getMessage(), e);
+            throw e;
```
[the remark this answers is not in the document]
JannalieT: removed try/catch

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/energy/RNARotamer.java
JannalieT: Changed to throw and IllegalArgumentException

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/energy/RotationalDynamics.java
JannalieT: Used Objects.requireNonNull

nmrfx-analyst/src/main/java/org/nmrfx/analyst/compounds/CompoundFitter.java
```diff
@@ -685,6 +688,7 @@ public double[] scoreByCMAES() {
                     new InitialGuess(start));
         } catch (Exception e) {
             e.printStackTrace();
+            throw e;
```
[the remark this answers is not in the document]
JannalieT: removed the try/catch

nmrfx-core/src/main/java/org/nmrfx/datasets/DatasetRegion.java
```diff
@@ -296,8 +300,18 @@ public int compareTo(Object o2) {
         return compare(this, o2);
     }
 
+    @Override
+    public int hashCode() {
+        int hash = 7;
+        hash = 31 * hash + (x == null ? 0 : Double.hashCode(x[0]));
+        return hash;
```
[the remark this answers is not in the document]
JannalieT: Changed to `return x == null ? Objects.hashCode(x) : Double.hashCode(x[0]);` since just using Objects.hashCode(x) alone failed the unittest checking if equal objects have the same hashcode, since the comparison is made on only the first element of x, not sure if this is okay?

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/OperationListCell.java
```diff
@@ -160,26 +160,17 @@ public void handle(DragEvent event) {
...
+                if (event != null && (target != null && target.getIndex() >= 0) ) {
```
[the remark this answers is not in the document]
JannalieT: removed extra parenthesis

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/ProcessorController.java
```diff
@@ -1011,23 +1011,21 @@ public void handle(ActionEvent e) {
...
             subMenuItems.add(menuItem);
         }
         menu.getItems().addAll(subMenuItems);
 
-        subMenuItems = null;
+        subMenuItems = new ArrayList<>();
         for (String op : OperationInfo.opOrders) {
             if (op.startsWith("Cascade-")) {
-                if (subMenuItems != null) {
+                if (!subMenuItems.isEmpty()) {
```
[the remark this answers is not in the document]
JannalieT: I don't think its necessary, removed the test

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/jeol/JeolDelta.java
```diff
@@ -576,32 +576,13 @@ int getInteger(byte[] values, int i) {
...
-                return null;
-            } else if (type.startsWith("T")) {
-                convertTime(values);
-                return null;
-            } else {
-                Number[] result = ByteConversion.convert(values, type, start, n);
-                return result;
-            }
+        Number[] convertNumber(byte[] values) {
+            return ByteConversion.convert(values, type, start, n);
```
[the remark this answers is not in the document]
JannalieT: The removed convertX methods that were called did not set any fields/do anything extra, their return values were not used and the method signature did not reflect their return values anyways so I think it is safe to remove them.

nmrfx-processor/src/main/java/org/nmrfx/processor/translate/Base64.java
```diff
@@ -229,44 +228,18 @@ public static String encodeObject(java.io.Serializable serializableObject) {
...
-        java.io.ObjectOutputStream oos = null;
-
-        try {
-            baos = new java.io.ByteArrayOutputStream();
-            b64os = new Base64.OutputStream(baos, Base64.ENCODE, breakLines);
-            oos = new java.io.ObjectOutputStream(b64os);
+                                      boolean breakLines) {
+        try (java.io.ByteArrayOutputStream baos = new java.io.ByteArrayOutputStream();
+             java.io.OutputStream b64os = new Base64.OutputStream(baos, Base64.ENCODE, breakLines);
+             java.io.ObjectOutputStream oos = new java.io.ObjectOutputStream(b64os)) {
```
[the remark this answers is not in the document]
JannalieT: used imports instead

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/energy/Dihedral.java
```diff
@@ -429,41 +429,24 @@ public void setBoundaries(final double sigma, boolean useDegrees, double maxRang
...
             } else {
-                inputSigma[aStart++] = sigma;
+                String parentName = atom.parent.getName();
+                double inputSigmaAtIndex;
+                if (Arrays.asList("P", "O5'", "C5'", "C4'", "O3'", "CA", "C").contains(parentName)) {
```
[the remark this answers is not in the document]
JannalieT: moved list outside of loop

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/predict/Predictor.java
```diff
@@ -485,7 +484,6 @@ public static void readFile(String resourceName) throws IOException {
                         if (fields.length > 0) {
                             if (fields[0].equals("rmax")) {
                                 setRMax(Double.parseDouble(fields[1]));
-                                aType = fields[2];
```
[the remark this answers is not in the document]
JannalieT: added a comment about what fields[2] is

nmrfx-structure/src/main/java/org/nmrfx/structure/rna/SSLayout.java
```diff
@@ -827,18 +827,15 @@ public PointValuePair refineCMAES(int nSteps, double stopFitness, final double s
...
             }
         }
-        if (result != null) {
-            //dumpCoordinates(result.getPoint());
-        } else {
+        if (result == null) {
             result = new PointValuePair(guess, value);
-            //dumpCoordinates(guess);
         }
         return result;
```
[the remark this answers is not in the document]
JannalieT: changed to ternary operator

nmrfx-structure/src/main/java/org/nmrfx/structure/rna/SSLayout.java
```diff
@@ -827,18 +827,15 @@ public PointValuePair refineCMAES(int nSteps, double stopFitness, final double s
                             new InitialGuess(lguess));
                 } catch (Exception e) {
                     e.printStackTrace();
+                    throw e;
```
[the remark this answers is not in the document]
JannalieT: removed try/catch

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/energy/CmaesRefinement.java
```diff
@@ -179,11 +179,11 @@ public double refineCMAESWithLinkedAtoms(final int nSteps, final double stopFitn
                     new InitialGuess(normBoundaries[2]));
         } catch (Exception e) {
             e.printStackTrace();
+            throw e;
```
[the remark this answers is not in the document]
JannalieT: removed try catch

nmrfx-structure/src/main/java/org/nmrfx/structure/seqassign/SpinSystems.java
JannalieT: Right okay, I will look into this one some more

nmrfx-core/src/main/java/org/nmrfx/datasets/DatasetRegion.java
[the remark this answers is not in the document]
vfiack: I missed the trap in the equals. Your change looks correct.
For info, Objects.hashCode(null) returns 0.

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/energy/Dihedral.java
```diff
@@ -425,17 +425,18 @@ public void setBoundaries(final double sigma, boolean useDegrees, double maxRang
...
+        List<String> parentNamesToScale = Arrays.asList("P", "O5'", "C5'", "C4'", "O3'", "CA", "C");
         for (int iAtom = 0; iAtom < angleAtoms.size(); iAtom++) {
             Atom atom = angleAtoms.get(iAtom).daughterAtom;
             // temporarily set everything to use sigma, till we make sure counting is correct
             if (atom == null) {
-                throw new NullPointerException("daughter null " + angleAtoms.get(iAtom).daughterAtom.getFullName());
+                throw new IllegalStateException("daughter null " + angleAtoms.get(iAtom).daughterAtom.getFullName());
             } else if (atom.parent == null) {
-                throw new NullPointerException("parent null " + angleAtoms.get(iAtom).daughterAtom.getFullName());
+                throw new IllegalStateException("parent null " + angleAtoms.get(iAtom).daughterAtom.getFullName());
```
vfiack, line 435: I know that this is exactly what I suggested, but if we decide to use requireNonNull() in some places, it may be a good idea to generalize it for all null checks? The main differences I see are the type of the thrown exception, and the time at which the name is created. 

Here the call to angleAtoms.get(..) is only done if atom is null, and it would be done all the time using requireNonNull(), so this version is slightly better performance-wise while the other is more readable. I'm honestly not sure which one we should prefer.

vfiack (commented):
Only the comment on SpinSystems is bothering me, everything else looks good to me. I'd still like if @brucejohnson could comment on the few comments where doubts remain (CompoundFitter and MoleculeBase mostly)

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/energy/Dihedral.java
```diff
@@ -425,45 +425,29 @@ public void setBoundaries(final double sigma, boolean useDegrees, double maxRang
...
-                inputSigma[aStart++] = sigma / backBoneScale;
-                if (sinCosMode) {
-                    inputSigma[aStart++] = sigma / backBoneScale;
-                }
-            } else if (atom.parent.getName().equals("CA") || atom.parent.getName().equals("C")) {
-                inputSigma[aStart++] = sigma / backBoneScale;
-                if (sinCosMode) {
-                    inputSigma[aStart++] = sigma / backBoneScale;
-                }
+                throw new IllegalStateException("daughter null " + angleAtoms.get(iAtom).daughterAtom.getFullName());
```
JannalieT, line 433: @brucejohnson suggestions for message? This is what was here before but I think that this message will result in a NPE for the getFullName() call as well since Atom atom = angleAtoms.get(iAtom).daughterAtom; and atom is null,

JannalieT, line 433: @brucejohnson suggestions for message? This is what was here before but I think that this message will result in a NPE for the getFullName() call as well since Atom atom = angleAtoms.get(iAtom).daughterAtom; and atom is null,

nmrfx-structure/src/main/java/org/nmrfx/structure/seqassign/SpinSystems.java
JannalieT: @vfiack throw a ParseException here instead as suggested by @brucejohnson

JannalieT: @vfiack throw a ParseException here instead as suggested by @brucejohnson

nmrfx-analyst/src/main/java/org/nmrfx/analyst/compounds/CompoundFitter.java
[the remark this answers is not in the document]
brucejohnson: Yes, I'd leave this for now.  The CompoundFitter code needs lots of work, but will definitely be relevant for some Nanalysis applications.

nmrfx-analyst/src/main/java/org/nmrfx/analyst/compounds/CompoundFitter.java
brucejohnson: A popup makes sense here, but again, this code needs a lot of work.  I'm not really happy with the whole interface / function here so some significant redesign may be warranted.

nmrfx-core/src/main/java/org/nmrfx/chemistry/Polymer.java
brucejohnson: We should throw an exception.  If atom1 or atom2 is null at this point something was wrong with list of atoms to constrain.

nmrfx-core/src/main/java/org/nmrfx/chemistry/io/PDBFile.java
brucejohnson: This is OK with your added test

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/FXMLController.java
brucejohnson: I don't think these will ever be null in normal usage so no message is needed to user.  I think the null checks and pritnlns  were put in during some initial testing where there was a problem in loading fxml files.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/jeol/JeolDelta.java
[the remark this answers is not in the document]
brucejohnson: Reading Jeol files is not fully implemented so I'm not sure if this code would ultimately be needed.  I started working on Jeol reading in collaboration with a friend who had Jeol instruments.  He moved universities (to one without Jeol) so I lost my supply of test data.  We need to get more files and ensure we can read them.

nmrfx-core/src/main/java/org/nmrfx/chemistry/Polymer.java
JannalieT: Used requireNonNull on the second attempt to set atom1 and atom2

nmrfx-analyst/src/main/java/org/nmrfx/analyst/compounds/CompoundFitter.java
JannalieT: Changed to use requireNonNull for the FitResult for now, so it is more obvious to add a popup here in the future if we decide to.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/jeol/JeolDelta.java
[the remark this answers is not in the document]
JannalieT: Added message explaining why code was removed as a reminder that it was removed for when JeolDelta is tested/improved in the future.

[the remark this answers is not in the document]
JannalieT: Added message explaining why code was removed as a reminder that it was removed for when JeolDelta is tested/improved in the future.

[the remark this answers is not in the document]
carlsonmark: test comment

vfiack (approved):

PR #122, opened 2022-05-20
NMR-4512: addedmethod to auto process 1D and 2D FID files, changed re…

JannalieT:
…turn type of loadDefaultScriptIfPresent method

### DIFF
--- a/nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/FXMLController.java
+++ b/nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/FXMLController.java
@@ -622,11 +622,27 @@ public class FXMLController implements  Initializable, PeakNavigable {
         if (!reload) {
             getActiveChart().full();
             getActiveChart().autoScale();
-            chartProcessor.loadDefaultScriptIfPresent();
+            generateScriptAndParse(nmrData);
         }
         getActiveChart().layoutPlotChildren();
         statusBar.setMode(0);
 
+    }
+
+    /**
+     * Check for existing default processing script and parse if present, otherwise generate
+     * and parse a processing script for FID data for 1D and 2D data.
+     * @param nmrData The NMRData set that will be processed.
+     */
+    private void generateScriptAndParse(NMRData nmrData) {
+        int nDim = nmrData.getNDim();
+        if (isFIDActive() && chartProcessor != null && !chartProcessor.loadDefaultScriptIfPresent() && (nDim == 1 || nDim == 2)) {
+            // TODO NMR-5184 update here if there is better way to determine if pseudo2D
+            // This is an estimate of whether the 2D data is pseudo2D, some pseudo2Ds may still be processed as 2Ds
+            boolean isPseudo2D = nmrData.getNVectors() < 100 && nDim == 2;
+            String script = chartProcessor.getGenScript(isPseudo2D);
+            processorController.parseScript(script);
+        }
     }
 
     public void addDataset(DatasetBase dataset, boolean appendFile, boolean reload) {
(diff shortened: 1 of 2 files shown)


nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/FXMLController.java
```diff
@@ -622,13 +622,29 @@ void addFID(NMRData nmrData, boolean clearOps, boolean reload) {
         if (!reload) {
             getActiveChart().full();
             getActiveChart().autoScale();
-            chartProcessor.loadDefaultScriptIfPresent();
+            generateScriptAndParse(nmrData);
```
JannalieT, line 625: The auto generate script will only happen if reload is false, but this is where the default script was checked for, so confirm if this is the correct spot?

vfiack: I think we should still load the script if it is present, and only generate if the script isn't there.

Think of this use case: 
1. You open nmrfx and a fid for the first time
2. This generates a default script (which is good, that's the change you made)
3. You change this script manually, click on process. This saves a process.py file in the FID directory
4. You close nmrfx and go back home
5. Next day, you open nmrfx. You want your changes from process.py to be loaded, not a new one being generated.

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/FXMLController.java
```diff
@@ -622,13 +622,29 @@ void addFID(NMRData nmrData, boolean clearOps, boolean reload) {
...
+     * Check for existing default processing script and parse if present, otherwise generate
+     * and parse a processing script for FID data for 1D and 2D data.
+     * @param nmrData The NMRData set that will be processed.
+     */
+    private void generateScriptAndParse(NMRData nmrData) {
+        int nDim = nmrData.getNDim();
+        if (isFIDActive() && chartProcessor != null && !chartProcessor.loadDefaultScriptIfPresent() && (nDim == 1 || nDim == 2)) {
+            // TODO NMR-5184 update here if there is better way to determine if pseudo2D
+            // This is an estimate of whether the 2D data is pseudo2D, some pseudo2Ds may still be processed as 2Ds
+            boolean isPseudo2D = nmrData.getNVectors() < 100 && nDim == 2;
```
vfiack, line 642: it would be nice to extract 100 as a constant, with a name such as `PSEUDO_2D_SIZE_THRESHOLD` (which is the one we use on spinlab)

vfiack (commented):
Tested, and this works as expected. A few minor comments.

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/FXMLController.java
JannalieT: Moved this value to a constant

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/FXMLController.java
```diff
@@ -622,13 +622,29 @@ void addFID(NMRData nmrData, boolean clearOps, boolean reload) {
...
     }
 
+    /**
+     * Check for existing default processing script and parse if present, otherwise generate
+     * and parse a processing script for FID data for 1D and 2D data.
+     * @param nmrData The NMRData set that will be processed.
+     */
+    private void generateScriptAndParse(NMRData nmrData) {
+        int nDim = nmrData.getNDim();
+        if (isFIDActive() && chartProcessor != null && !chartProcessor.loadDefaultScriptIfPresent() && (nDim == 1 || nDim == 2)) {
```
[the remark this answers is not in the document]
JannalieT: added info log messages to this method and to the loadDefaultScriptIfPresent method

brucejohnson (approved):

PR #125, opened 2022-06-01
Nmr 5224 log console basic features

JannalieT:


nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/log/LogConsoleController.java
```diff
@@ -0,0 +1,213 @@
...
+    private MasterDetailPane logDisplayMasterDetail;
+
+    private final LogTable table = new LogTable();
+    private final LogDetailsView logDetails = new LogDetailsView();
+
+    /**
+     * Creates a new LogConsoleController and sets the static LogConsoleController.
+     * @return The newly created LogConsoleController
+     */
+    public static LogConsoleController create() {
```
JannalieT, line 52: Tried to follow the same pattern as the ConsoleController

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/log/LogLevelCellFactory.java
```diff
@@ -0,0 +1,45 @@
...
+    public static Glyph selectIcon(LogLevel logLevel) {
+        int intValue = logLevel.getLogbackLevel().toInteger();
+        if (intValue >= Level.ERROR_INT)
+            return new Glyph(FONT_FAMILY, FontAwesomeIcon.CLOSE).color(Color.RED);
+        if (intValue >= Level.WARN_INT)
+            return new Glyph(FONT_FAMILY, FontAwesomeIcon.WARNING).color(Color.GOLD);
+        if (intValue >= Level.INFO_INT)
+            return new Glyph(FONT_FAMILY, FontAwesomeIcon.INFO_CIRCLE).color(Color.DEEPSKYBLUE);
+        if (intValue >= Level.DEBUG_INT)
+            return new Glyph(FONT_FAMILY, "d").color(Color.GREY);
```
JannalieT, line 41: Just used the letter 'd' for the debug icon. The other option could be to use a 'bug' icon here?

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/log/LogDetailsView.java
```diff
@@ -0,0 +1,92 @@
...
+
+    private static final int SPACING = 5;
+    private final TextField datetime = new TextField();
+    private final TextField location = new TextField();
+    private final TextArea message = new TextArea();
+    private final Button copyButton = GlyphsDude.createIconButton(FontAwesomeIcon.COPY);
+    private String level;
+
+    public LogDetailsView() {
+        super();
```
vfiack, line 34: default constructor call to super is implicit in java, you don't need to write it.
```suggestion
```

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/log/LogDetailsView.java
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Sets the controls with the values from logRecord.
+     * @param logRecord The LogRecord object.
+     */
+    public void setDetails(LogRecord logRecord) {
+        if (logRecord == null) {
+            level  = null;
+            datetime.setText("");
+            location.setText("");
+            message.setText("");
```
vfiack, line 71: I think it would look a bit better to extract this as a `clearDetails()` method or similar, and call it directly from the constructor as well to avoid passing null parameters.

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/log/LogLevelCellFactory.java
```diff
@@ -0,0 +1,45 @@
+package org.nmrfx.processor.gui.log;
+
+import de.jensd.fx.glyphs.fontawesome.FontAwesomeIcon;
+import javafx.scene.control.TableCell;
+import javafx.scene.paint.Color;
+import org.controlsfx.glyphfont.Glyph;
+
+import ch.qos.logback.classic.Level;
```
vfiack, line 8: I had hoped to abstract the dependency to logback. It isn't a good news to see this import in UI code. 
Is there any way to use the `org.nmrfx.processor.gui.log.LogLevel` enum here?

If it is only for the `>=` checks, we could use `LogLevel.ordinal()` instead.

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/log/LogDetailsView.java
```diff
@@ -0,0 +1,92 @@
...
+ * Class to display the details of a LogRecord.
+ */
+public class LogDetailsView extends GridPane {
+
+    private static final int SPACING = 5;
+    private final TextField datetime = new TextField();
+    private final TextField location = new TextField();
+    private final TextArea message = new TextArea();
+    private final Button copyButton = GlyphsDude.createIconButton(FontAwesomeIcon.COPY);
+    private String level;
```
vfiack, line 31: Why using the level as string? It looks like it's only usage is to copy to clipboard, so we could store the LogLevel or even the LogRecord directly here, and convert to string only when copying to clipboard.

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/log/LogTimeCellFactory.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+
+
+/**
+ * Formatter for LocalDateTime table cells in a LogRecord table
+ */
+public class LogTimeCellFactory extends TableCell<LogRecord, LocalDateTime> {
```
vfiack, line 13: It is strange to have a `SomethingFactory` extends `Something`. Is it really a factory?

nmrfx-processor-gui/src/test/java/org/nmrfx/processor/gui/log/LogConsoleControllerTest.java
```diff
@@ -0,0 +1,159 @@
...
+import junit.framework.TestCase;
+import org.junit.Before;
+import org.junit.Test;
+import org.slf4j.Marker;
+
+import java.util.ArrayList;
+import java.util.Map;
+import java.util.function.Predicate;
+
+public class LogConsoleControllerTest extends TestCase {
```
vfiack, line 19: extending TestCase is obsolete, and comes from a time before `@Test` annotations. It shouldn't be necessary.

nmrfx-processor-gui/src/test/java/org/nmrfx/processor/gui/log/LogConsoleControllerTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+    private static class LoggingEventTest implements ILoggingEvent {
```
vfiack, line 21: Wouldn't this be better as a mock? (see Mockito, we can provide examples from spinlab as well).
Otherwise, you could maybe extend `ch.qos.logback.classic.spi.LoggingEvent` directly to avoid having all these empty methods.

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/log/LogDetailsView.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /**
+     * Copies the values of the controls as well as the software version to the
+     * system clipboard.
+     * @param actionEvent
+     */
+    private void toClipboardClicked(ActionEvent actionEvent) {
+        String content = "[" + level + ":" + datetime.getText() + "] " + location.getText() + "\n" + message.getText()
+                + "\n\nSoftware version: " + NvUtil.getVersion();
```
vfiack, line 87: This returns null when running from IDE. Not a big deal, but could maybe handle the case and write "development" or something.

vfiack (commented):
Working as expected 👍 

A few improvement suggestions.

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/log/LogTimeCellFactory.java
JannalieT: I thought it was because I was using it to set the cellFactoryProperty but looking closer at the example in the docs they just called their class SomethingFormat. I will rename to LogTimeCellFormatter.

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/log/LogDetailsView.java
JannalieT: removed this one and the other one in LogTable

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/log/LogDetailsView.java
JannalieT: made clearDetails method

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/log/LogLevelCellFactory.java
JannalieT: Was able to use LogLevel.ordinal instead, removed dependency to logback

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/log/LogDetailsView.java
JannalieT: Added "development" in NVUtils, not sure if thats okay or if I should only handle the case in LogDetailsView, now when running from IDE, then the window title will also say development instead of NULL
image

also fixed NPE here, by copying empty string to clipboard if no log selected

vfiack (approved):

PR #127, opened 2022-06-07
Nmr 5109 replace syserr with loggers

JannalieT:
Replaced System.err with log messages, and replaced any System.out in files with System.err with log messages
Switched some System.err messages to use log.info instead of log.warn (see comment for example)
https://github.com/nanalysis/nmrfx/pull/127/files#r891760910
Switched some System.out to use log.warn instead of log.info, if the message sounds like something failed (see comment for example)
https://github.com/nanalysis/nmrfx/pull/127/files#r891761126

Used log.debug for print statements in dumpX style methods (see comment for example) 
https://github.com/nanalysis/nmrfx/pull/127/files#r891763718

### DIFF
--- a/nmrfx-core/src/main/java/org/nmrfx/datasets/SubMatrixFile.java
+++ b/nmrfx-core/src/main/java/org/nmrfx/datasets/SubMatrixFile.java
@@ -75,10 +75,10 @@ public class SubMatrixFile implements DatasetStorageInterface, Closeable {
         fc = raFile.getChannel();
         int blockHeaderSize = layout.getBlockHeaderSize() / BYTES;
         long matSize = BYTES;
-        System.err.println(dataset.getFileName());
-        System.err.println("header size " + layout.getFileHeaderSize());
+        log.info(dataset.getFileName());
+        log.info("header size {}", layout.getFileHeaderSize());
         for (int i = 0; i < dataset.getNDim(); i++) {
-            System.err.println("sub cache " + i + " " + layout.blockSize[i] + " " + layout.nBlocks[i] + " " + dataset.getSizeTotal(i));
+            log.info("sub cache {} {} {} {}", i, layout.blockSize[i], layout.nBlocks[i], dataset.getSizeTotal(i));
             matSize *= (layout.blockSize[i] + blockHeaderSize) * layout.nBlocks[i];
         }
         totalSize = matSize / BYTES;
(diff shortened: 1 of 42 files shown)


nmrfx-core/src/main/java/org/nmrfx/datasets/SubMatrixFile.java
```diff
@@ -75,10 +75,10 @@ void init() throws IOException {
         fc = raFile.getChannel();
         int blockHeaderSize = layout.getBlockHeaderSize() / BYTES;
         long matSize = BYTES;
-        System.err.println(dataset.getFileName());
-        System.err.println("header size " + layout.getFileHeaderSize());
+        log.info(dataset.getFileName());
+        log.info("header size {}", layout.getFileHeaderSize());
```
JannalieT, line 79: Used log.info here since its informative and nothing bad seems to be happening

nmrfx-core/src/main/java/org/nmrfx/datasets/SubMatrixFile.java
```diff
@@ -178,20 +174,18 @@ public long getTotalSize() {
...
     synchronized ByteBuffer readBlock(long iBlock) throws IOException {
-//        System.out.println("read block " + iBlock);
         long blockPos = iBlock * (layout.blockPoints * BYTES + layout.blockHeaderSize) + layout.fileHeaderSize;
         ByteBuffer buffer = ByteBuffer.allocate((int) (layout.blockPoints * BYTES));
         buffer.order(dataset.getByteOrder());
 
         int nc = fc.read(buffer, blockPos);
         if (nc != layout.blockPoints * BYTES) {
-            System.out.println("read failed " + nc);
+            log.warn("read failed {}", nc);
```
JannalieT, line 183: Used log.warn here since the message says the read failed

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/jeol/JeolDelta.java
```diff
@@ -842,8 +841,12 @@ void loadParams(int parStart, int parLength) {
...
 
     public void dumpPars() {
-        for (String key : parMap.keySet()) {
-            System.out.println(parMap.get(key));
+        if (log.isDebugEnabled()) {
+            StringBuilder parString = new StringBuilder();
+            for (String key : parMap.keySet()) {
+                parString.append(parMap.get(key)).append(System.lineSeparator());
+            }
+            log.debug(parString.toString());
```
JannalieT, line 849: Switched to use log.debug

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/BigMappedMatrixFile.java
```diff
@@ -275,7 +273,7 @@ public double sumValues() throws IOException {
                 sum += getMapping(mapN).getFloat(offN);
             } catch (IOException e) {
                 MappedByteBuffer mapping = getMapping(mapN);
-                System.out.println(mapN + " Err " + offN + " " + mapping.capacity() + " " + mapping.limit());
+                log.error("{} Err {} {} {}", mapN, offN , mapping.capacity(), mapping.limit());
                 System.exit(0);
```
JannalieT, line 277: Changed to log.error since immediately followed by System.exit....the exit code is 0 though, but this isn't a normal exit

vfiack (approved):

PR #129, opened 2022-06-15
NMR-5222: add nmrfx-test-data repository as submodule and use Tempora…

JannalieT:
…ryFolders for writing test files

### DIFF
--- a/nmrfx-processor/src/test/java/org/nmrfx/processor/datasets/ProcessTest.java
+++ b/nmrfx-processor/src/test/java/org/nmrfx/processor/datasets/ProcessTest.java
@@ -13,12 +13,14 @@ import java.io.IOException;
 import java.nio.file.FileSystems;
 import java.nio.file.Path;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assume.assumeFalse;
 
 public class ProcessTest {
 
     private static final String VALID_SUBMODULE_LOCATION = "nmrfx-test-data/valid";
     private static final String FID_SUBMODULE_LOCATION = "nmrfx-test-data/testfids/";
+    private static final String ERR_MSG = "File doesn't exist: ";
     public static String scriptHome = "src/test/resources/process_scripts";
     private static String fidHome;
     private static String tmpHome;
@@ -57,7 +59,7 @@ public class ProcessTest {
         if (!testFile.exists()) {
             boolean isBuildEnv = Boolean.parseBoolean(System.getenv("BUILD_ENV"));;
             if (isBuildEnv) {
-                throw new FileNotFoundException("Missing build environment requirement. " +"File doesn't exist." + testFile);
+                throw new FileNotFoundException("Missing build environment requirement. " + ERR_MSG + testFile);
             }
             return true;
         } else {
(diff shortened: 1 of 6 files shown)


nmrfx-processor/src/test/java/org/nmrfx/processor/datasets/ProcessTest.java
```diff
@@ -1,18 +1,48 @@
...
-    public static String validHome = "../../nmrfxp_tests/valid";
+    private static String fidHome;
+    private static String tmpHome;
+    private static String validHome;
+
+    @ClassRule
+    public static final TemporaryFolder tmpFolder = TemporaryFolder.builder()
+            .parentFolder(new File(System.getProperty("user.dir")))
+            .assureDeletion()
+            .build();
```
JannalieT, line 33: tests will fail if temporary folder is not deleted, I can also change it to pass the tests without checking if the folder is actually deleted if that is preferred

nmrfx-processor/src/test/java/org/nmrfx/processor/datasets/ProcessTest.java
```diff
@@ -1,18 +1,48 @@
...
+    @BeforeClass
+    public static void setup() {
+        Path parent = FileSystems.getDefault()
+                .getPath("")
+                .toAbsolutePath()
+                .getParent();
+        // Python scripts expect path with / not \, so replace if present
+        fidHome = parent.resolve(FID_SUBMODULE_LOCATION).toString().replace("\\", "/") + "/";
+        validHome = parent.resolve(VALID_SUBMODULE_LOCATION).toString().replace("\\", "/") +"/";
+        tmpHome = tmpFolder.getRoot().toString().replace("\\", "/") + "/";
```
JannalieT, line 44: Was trying to avoid having to use `../../` in a hardcoded string and have the path automatically created, but had issues with the python scripts expected file separators. Not sure if this is any better than the hardcoded strings.

nmrfx-processor/src/test/java/org/nmrfx/processor/datasets/ProcessTest.java
```diff
@@ -32,133 +73,134 @@ public long runAndCompare(String fileName) throws IOException {
...
         // so, currently,  if file doesn't exist a message is printed and the test passes
-        if (!refFile.exists()) {
-            System.out.println(refFile + " doesn't exist, skipping test");
-            return 0;
-        } else {
-            executeScript(fileName);
-            long result = DatasetCompare.compare(refFile, testFile);
-            return result;
-        }
+        assumeFalse(ERR_MSG + refFile, testFilesMissing(refFile));
```
JannalieT, line 86: If a user without files runs the tests, they will show up as skipped rather than passed with a message. I think this is better because its a little more obvious that user hasn't fully been able to run the tests.

JannalieT:
There will be still one more small change after the nmrfx-test-data branch is merged, I will update the submodule commit.

Also, the tests take approximately 5 times as long to run now that we are not skipping the 25 tests.

brucejohnson (dismissed):

PR #132, opened 2022-06-23
specify classifiers for platform-specific dependencies

vfiack:
also fix javacpp/openblas version mismatch

This should generate smaller assembly and maybe build a little faster. 
The main goal is to remove the need for `cleanlibs` script before generating installers.

I would have liked to be able to only exclude certain classifiers, but maven doesn't seem to have this feature, so we have to exclude the main dependency and include manually the specific platforms we need.

**Edit after build**: the analyst zip assembly weighs 250Mb less!

### DIFF
--- a/nmrfx-bom/pom.xml
+++ b/nmrfx-bom/pom.xml
@@ -15,6 +15,10 @@
     <properties>
         <nd4j.version>1.0.0-beta7</nd4j.version>
         <dl4j.version>1.0.0-beta7</dl4j.version>
+        <openblas.version>0.3.20-1.5.8-SNAPSHOT</openblas.version>
+        <arpack.version>3.8.0-1.5.7</arpack.version>
+        <javacpp.version>1.5.8-SNAPSHOT</javacpp.version>
+        <mkl.version>2020.1-1.5.3</mkl.version>
     </properties>
 
     <dependencyManagement>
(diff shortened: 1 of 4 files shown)


JannalieT (approved):

nmrfx-bom/pom.xml
```diff
@@ -15,6 +15,10 @@
     <properties>
         <nd4j.version>1.0.0-beta7</nd4j.version>
         <dl4j.version>1.0.0-beta7</dl4j.version>
+        <openblas.version>0.3.20-1.5.8-SNAPSHOT</openblas.version>
+        <arpack.version>3.8.0-1.5.7</arpack.version>
+        <javacpp.version>1.5.8-SNAPSHOT</javacpp.version>
```
vfiack, line 20: @brucejohnson just to double check: this was previously using 1.5.7, but had the native builds from snapshot transitively from openblas.

nmrfx-bom/pom.xml
```diff
@@ -112,7 +116,71 @@
...
+                <artifactId>nd4j-native</artifactId>
+                <version>${nd4j.version}</version>
+                <classifier>linux-x86_64</classifier>
+                <!-- license: Apache License, Version 2.0 -->
+            </dependency>
+            <dependency>
+                <groupId>org.nd4j</groupId>
+                <artifactId>nd4j-native</artifactId>
+                <version>${nd4j.version}</version>
+                <classifier>macosx-x86_64</classifier>
```
vfiack, line 141: There doesn't seem to be a build for macosx-arm64 for nd4j, mkl or arpack-ng. Is that a problem?

brucejohnson: It's a pain that these projects haven't yet updated to having arm64, as Mac's with them have been available for quite a while now.  I'll test this out.  I've got Macs with and without arm64 now so I can make sure it all runs.

I'm not sure if there will ever be MKL for arm, as MKL is from Intel

brucejohnson (changes requested):
I tested out an operation that uses deeplearning4j (predicting protein secondary structure) and it fails.  After some testing I found that we need to have
nd4j-native-1.0.0-beta7.jar included in the build, but it is getting excluded by:
                    <exclusion>
                        <groupId>org.nd4j</groupId>
                        <artifactId>nd4j-native</artifactId>
                    </exclusion>

That's there to remove the platform specific files that we don't need, but also excludes this one that we do.

I'm working on a JUnit test that can make sure the deep learning code works.

vfiack:
>  After some testing I found that we need to have nd4j-native-1.0.0-beta7.jar included in the build

This one is problematic, they use the same artefact name for the "api" part and for the native implementations, and maven doesn't seem to support classifier-only exclusions. One workaround would be to not exclude it at all, but then we need the cleanup script or equivalent, or ship all platforms.

vfiack:
Let's close this one and try https://github.com/nanalysis/nmrfx/pull/139 instead.

PR #133, opened 2022-06-24
Nmr 5167 opening fid with grid display

JannalieT:
Major Pr Changes:
- Moved 2D processing widgets (slider and radio buttons) to the processor controller under a new tab called "Navigation" and removed the "Options" tab
- Allow multiple FIDs to be loaded
- Changing the active chart changes the spectrum status bar tool options

Some of the grid functionality is slightly different than before due to allowing the multiple processing controllers so will need some extensive testing to make sure it still has the desired functionality.
Note: was unable to test changes with Scanner Table very well, as the functionality before this pr was not what I expected, will need to confirm the behaviour and likely create a jira issue for it.


JannalieT:
There was an NMRScene.fxml file randomly outside of resources, just inside the nmrfx-processor-gui directory, I removed it since it didn't seem like it was being used.

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/FXMLController.java
```diff
@@ -112,47 +112,27 @@ public class FXMLController implements  Initializable, PeakNavigable {
...
     ChartProcessor chartProcessor;
     DocWindowController dwc = null;
     static SpecAttrWindowController specAttrWindowController = null;
     static boolean popOverMode = false;
     static PeakAttrController peakAttrController = null;
-    ProcessorController processorController = null;
     Stage stage = null;
+    private double previousStageRestoreWidth = 0;
+    private double previousStageRestoreProcControllerWidth = 0;
+    private boolean previousStageRestoreProcControllerVisible = false;
```
JannalieT, line 135: Allowing multiple FID processor controllers required setting the size in a different way, because the previous sizing method would continuously increase the stage size each time a new FID was loaded. The size is now adjusted in two methods: 

- adjustSizeAfterMaximize: Uses these three variables to adjust the size when restoring down after the window has been maximized. I don't really like this solution but I couldn't come up with anything else...any suggestions would be welcome.
- updateStageSize : Adjusts the size of the window when loading an FID or dataset

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/FXMLController.java
```diff
@@ -696,6 +679,28 @@ public void addDataset(DatasetBase dataset, boolean appendFile, boolean reload)
         getActiveChart().layoutPlotChildren();
     }
 
+    /**
+     * Updates the SpectrumStatusBar menu options based on whether FID mood is on, and the dimensions
+     * of the dataset.
+     */
+    private void updateSpectrumStatusBarOptions() {
+        if (isFIDActive()) {
```
JannalieT, line 687: When in grid display, when selecting a 1D vs 2D vs FID etc graph the tool bar options on the bottom should update.

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/ProcessorController.java
```diff
@@ -1126,6 +1137,29 @@ public void changed(ObservableValue<? extends Number> observableValue, Number nu
...
+
+    }
+
+    /**
+     * Listener for the Processor availability status and updates the processor available status
+     * @param newStatus the newly updated status
+     */
+    public void processorAvailableStatusUpdated(Boolean newStatus) {
+        processorAvailable.setValue(newStatus);
+    }
```
JannalieT, line 1156: Used observer pattern to disable "Process" button when mutliple processor controllers were created

brucejohnson:
The Processor menu item should be disabled if the active window doesn't have an FID associated with it.  Currently the user can select the item and nothing happens if they are on a dataset window.

brucejohnson:
If you close the processor sidebar with the close icon (top right) and then click on the window the side bar comes back.  If the user wants to hide it, then it should only return if you select "Show Processor" from the View menu.  Otherwise there is no way to keep it hidden (at least as far as I can tell)

JannalieT:
> If you close the processor sidebar with the close icon (top right) and then click on the window the side bar comes back. If the user wants to hide it, then it should only return if you select "Show Processor" from the View menu. Otherwise there is no way to keep it hidden (at least as far as I can tell)

changed to only show the processor if the processor hasn't been hidden

brucejohnson (dismissed):

brucejohnson (approved):

PR #134, opened 2022-06-28
Nmr 5107 remove useless loops

JannalieT:
- removed 3 useless loops in SpatialSetGroup, Align and Molecule
- replaced System.out with log in all files affected by the above point's changes

### DIFF
--- a/nmrfx-analyst/src/main/java/org/nmrfx/analyst/dataops/Align.java
+++ b/nmrfx-analyst/src/main/java/org/nmrfx/analyst/dataops/Align.java
@@ -47,13 +47,13 @@ public class Align {
 
             dataset.readVector(vec2, j, 0);
             double corr = compareVec(vec1, vec2, pt1, pt2);
-            System.out.println(row + " " + j + " " + corr);
+            log.info("{} {} {}", row, j, corr);
             if (corr > maxCorr) {
                 maxCorr = corr;
                 index = j;
             }
         }
-        System.out.println(row + " max " + index + " " + maxCorr);
+        log.info("{} max {} {}", row, index, maxCorr);
         return index;
     }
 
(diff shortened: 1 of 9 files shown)


nmrfx-analyst/src/main/java/org/nmrfx/analyst/dataops/Align.java
```diff
@@ -439,35 +438,21 @@ public void segmentedAlign(Vec targetVec, Vec sampleVec, Vec resultVec, int preE
...
-//                if (maxShiftA <  0) {
-//                    break;
-//                }
-            }
-//            if (shift < 0) {
-//                shift = 0;
-//            }
-            Alignment alignment = new Alignment(region.min, region.max, pt1e + shift, pt1e + shift + size - 1, shift, shift - preExtra, corrValue, ratio, useSection);
+            // Note: originally this method tried three times to calculate the shift to get a shift that was greater
+            // than 0 and then used a shift of 0 if the shift was still <0 but the code was commented out so removed it
```
JannalieT, line 443: Removed commented out code, but left note to explain that there used to be some commented out code/for looop

nmrfx-structure/src/main/java/org/nmrfx/structure/noe/NOECalibrator.java
```diff
@@ -422,10 +415,25 @@ public void findRedundant() {
...
+     * @return The entity cast to a residue
+     */
+    private Residue checkEntityIsResidue(Entity entity) {
+        Residue residue = null;
+        if (entity instanceof Residue ) {
+            residue = (Residue) entity;
+        } else {
+            log.info("Entity is not a polymer: {}", entity.getName());
+        }
+        return residue;
```
vfiack, line 430: Minor, but this could be slightly simplified:

```suggestion
        if (entity instanceof Residue) {
            return (Residue) entity;
        } else {
            log.info("Entity is not a polymer: {}", entity.getName());
            return null;
        }
```

That way it's clearer that the intermediate variable is never accessed.

vfiack (dismissed):

PR #135, opened 2022-06-29
Nmr 5329 save more order parameter info

brucejohnson:


nmrfx-core/src/main/java/org/nmrfx/chemistry/relax/OrderPar.java
```diff
@@ -287,11 +286,11 @@ public Double getValue() {
...
         return error;
     }
-//    static final String[] PAR_NAMES = {"Order_param", "Tau_e", "Tau_f", "Tau_s", "Rex", "Sf2", "Ss2"};
 
     @Override
     public Double getValue(String name) {
         switch (name) {
+            case "Order_param":
             case "S2":
                 return value;
```
JannalieT, line 295: If "Order_param" and "S2" are both returning the same value I think you can write it like this
```suggestion
            case "Order_param", "S2":
                return value;
```

nmrfx-core/src/main/java/org/nmrfx/chemistry/relax/OrderPar.java
```diff
@@ -193,6 +187,7 @@ public OrderPar set(String name, Double val, Double err) {
                 Ss2, Ss2err,
                 sumSqErr, nValues, nPars, model);
         switch (name) {
+            case "Order_param":
             case "S2":
                 newPar.value = val;
                 newPar.error = err;
```
JannalieT, line 193: If "Order_param" and "S2" are supposed to be doing the same thing I think you can write it like this:
```suggestion
            case "Order_param", "S2":
                newPar.value = val;
                newPar.error = err;
```

nmrfx-core/src/main/java/org/nmrfx/chemistry/relax/OrderPar.java
```diff
@@ -325,9 +326,30 @@ public Double getValue(String name) {
...
+            return 0.0;
+        }
+    }
+
     @Override
     public Double getError(String name) {
         switch (name) {
+            case "Order_param":
             case "S2":
                 return error;
```
JannalieT, line 354: see above comments
```suggestion
            case  "Order_param", "S2":
                return error;
```

nmrfx-core/src/main/java/org/nmrfx/chemistry/relax/OrderPar.java
```diff
@@ -398,24 +421,63 @@ public String toString() {
         RelaxationValues.appendValueError(sBuilder, TauS, TauSerr,"%.4f");
         RelaxationValues.appendValueError(sBuilder, Rex, Rexerr,"%.2f");
         sBuilder.append("\t").append(model).append("\t").append(modelNum);
+        sBuilder.append("\t").append(String.format("%.4f", sumSqErr)).append("\t").
+                append(String.format("%.4f", getReducedChiSqr())).
+                append("\t").append(String.format("%.4f",getAIC())).append("\t").
+        append(nValues).append("\t").append(nPars);
```
JannalieT, line 427: This is more of a style thing, but I find it easier to read when the dots are on the start of a row
```suggestion
        sBuilder.append("\t").append(String.format("%.4f", sumSqErr)).append("\t")
                .append(String.format("%.4f", getReducedChiSqr()))
                .append("\t").append(String.format("%.4f",getAIC())).append("\t")
                .append(nValues).append("\t").append(nPars);
```

JannalieT (commented):
I think some of the sonar lint issues should be fixed, there are a couple unused variables that were left in after some of the code deletions

nmrfx-core/src/main/java/org/nmrfx/chemistry/relax/OrderPar.java
JannalieT: Oh actually this comma separated format was only available from java 14 on, so I don't know if we can make this change yet?

nmrfx-core/src/main/java/org/nmrfx/chemistry/relax/OrderPar.java
brucejohnson: That is a style thing and I don't know if there is a standard for this.  I like the dots at the end because when I read the line to the end and see a dot I know that it is continued on the next line.

nmrfx-core/src/main/java/org/nmrfx/chemistry/relax/OrderPar.java
brucejohnson: Yes, I haven't started moving to post 11 features

nmrfx-core/src/main/java/org/nmrfx/chemistry/relax/OrderPar.java
brucejohnson: I've cleaned up some of the sonar hints

JannalieT (approved):

PR #137, opened 2022-06-29
Nmr 5264 generate prf files for unnatural amino acids

brucejohnson:


nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/energy/AngleTreeGenerator.java
```diff
@@ -426,56 +437,51 @@ public void measureAtomTree(ITree itree, List<List<Atom>> atomTree) {
...
+            for (int i=branch.size();i<8;i++) {
+                System.out.printf("%8s", "");
+            }
+            Atom branchAtom = branch.get(2);
+            System.out.printf(" %3d %6b %3d ", branchAtom.rotUnit, branchAtom.rotActive, branchAtom.irpIndex);
+            for (int i = 3;i<branch.size();i++) {
+                branchAtom = branch.get(i);
+                System.out.printf("%6s %8.2f %8.2f %8.2f",branchAtom.getName(), Math.toDegrees(branchAtom.valanceAngle), branchAtom.bondLength, Math.toDegrees(branchAtom.dihedralAngle));
+            }
+            System.out.println();
```
JannalieT, line 484: we should change these print statements to logs since we are trying to remove the print statements

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/energy/AngleTreeGenerator.java
```diff
@@ -524,6 +532,12 @@ private static void addRingClosurePairs(Map<Atom, Map<Atom, Double>> ringClosure
         }
     }
 
+    public static void addConstrainDistance(Map<Atom, Map<Atom, Double>> ringClosures, Atom begin, Atom end) {
+        addRingClosure(ringClosures, begin, end);
+    }
+
+
+
     public List<Atom> getPathList() {
```
JannalieT, line 541: too many spaces
```suggestion
    }

    public List<Atom> getPathList() {
```

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/miner/PathIterator.java
```diff
@@ -267,23 +258,21 @@ void dumpPath() {
...
             System.out.print(iAtom + " ");
         }
-        System.out.println("");
+        System.out.println();
         for (int i = 0; i < n; i++) {
             IAtom atom = ac.getAtom(getAtomIndex(i));
             System.out.print(atom.getSymbol());
         }
-        System.out.println("");
+        System.out.println();
```
JannalieT, line 266: These should probably be changed to log.debug, you can use String builder and log just once at the end

nmrfx-structure/src/main/resources/refine.py
```diff
@@ -2487,12 +2502,19 @@ def setupTree(self, treeDict):
                                  treeDict['end'] if 'end' in treeDict else None)
                                  if treeDict else (None, None))
 
+        print 'tree',start,end
```
JannalieT, line 2505: can this be removed?

JannalieT (commented):
Is it possible to add a unit test so that we can check that the prf files are generated correctly?

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/energy/AngleTreeGenerator.java
brucejohnson: Switched to logs (though still a method which uses StringBuilder and can return string, I call this from Python while working with new structures)

nmrfx-structure/src/main/java/org/nmrfx/structure/chemistry/miner/PathIterator.java
brucejohnson: Switched to logs (though still a method which uses StringBuilder and can return string, I call this from Python while working with new structures)

nmrfx-structure/src/main/resources/refine.py
brucejohnson: removed

brucejohnson:
Despite the name, these changes are really infrastructure for generating the prf files.  The real work gets done in code that is in nmrfx-private.  That's where I should put tests for prf generation.  I've done a lot of testing with a fairly complex molecule, but it is a proprietary compound from a customer.  I need to find one with similar complexity that is not proprietary.  When I do I'll add a test.

JannalieT (approved):
Noticed a few log messages had str + str in them, you can actually write them like this `log.info("value of varA: {}, value of varB: {}", varA, varB);` and its supposed to be a bit better performance wise but the actual output is the same

PR #138, opened 2022-06-29
Nmr 5331 add peak prediction for 1 d peaks

brucejohnson:


nmrfx-structure/src/main/java/org/nmrfx/structure/tools/PeakGenerator.java
```diff
@@ -0,0 +1,230 @@
...
+                });
+    }
+
+    public void generateHSQC(PeakList peakList, int parentElement) {
+        Atom[] atoms = new Atom[2];
+        molecule.getAtoms().stream().
+                filter(atom -> atom.getAtomicNumber() == 1).
+                filter(atom -> !atom.isMethyl() || atom.isFirstInMethyl()).
+                filter(atom -> atom.getParent() != null && atom.getParent().getAtomicNumber() == parentElement).
+                forEach(atom -> {
```
JannalieT, line 223: same here if you want to change
```suggestion
        molecule.getAtoms().stream()
                .filter(atom -> atom.getAtomicNumber() == 1)
                .filter(atom -> !atom.isMethyl() || atom.isFirstInMethyl())
                .filter(atom -> atom.getParent() != null && atom.getParent().getAtomicNumber() == parentElement)
                .forEach(atom -> {
```

JannalieT (commented):
This looks okay, but do the new methods in PeakGenerator get used anywhere?

nmrfx-structure/src/main/java/org/nmrfx/structure/tools/PeakGenerator.java
brucejohnson: The new method is called from a class in nmrfx-private, so it isn't seen in this repository

JannalieT (dismissed):

JannalieT (approved):

PR #140, opened 2022-07-05
NMR-5223: generate installer automatically

vfiack:
Call installer creation and upload workflow from private repository.

Some open questions remains:
* do we build the installer on every PR's push? or only on master or release branches?
* do we keep the 'basic' build archive as well?
* do we keep the cleanup in the public repo? or move it to the private one as well to remove all accesses to sharepoint from the public repo?


.github/workflows/build.yml
```diff
@@ -81,20 +81,19 @@ jobs:
...
-          SHAREPOINT_CLIENT_SECRET: ${{ secrets.SHAREPOINT_CLIENT_SECRET }}
-        run: bash .github/workflows/actions-publish-build.sh
\ No newline at end of file
+          github-token: ${{ secrets.PRIVATE_REPO_TOKEN }}
+          script: |
+            await github.rest.actions.createWorkflowDispatch({
+              owner: 'nanalysis',
+              repo: 'nmrfx-private',
+              workflow_id: 'build-installers.yml',
+              ref: 'NMR-5223-create-installer-on-trigger',
```
vfiack, line 93: This will need to be replaced with `master` once the private repo's branch is merged.

vfiack: Done.

JannalieT (approved):

PR #142, opened 2022-07-08
NMR-4812: load processed JCAMP data into memory file datasets

JannalieT:


nmrfx-core/src/main/java/org/nmrfx/project/ProjectBase.java
```diff
@@ -145,11 +147,10 @@ List<DatasetBase> getDatasetList() {
 
     public List<DatasetBase> getDatasetsWithFile(File file) {
         try {
-            String testPath = file.getCanonicalPath();
-            List<DatasetBase> datasetsWithFile = datasetMap.values().stream().filter((dataset) -> (dataset.getCanonicalFile().equals(testPath))).collect(Collectors.toList());
-            return datasetsWithFile;
+            String testPath = requireNonNullElse(file.getCanonicalPath(), "");
+            return datasetMap.values().stream().filter(dataset -> (testPath.equals(dataset.getCanonicalFile()))).collect(Collectors.toList());
```
JannalieT, line 151: Since I used the in memory file, getCanonicalFile was returning null and causing an exception, changed to default to an empty string. I'm not sure if I should have tried comparing the dataset names instead if they are memory files?

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/Dataset.java
```diff
@@ -321,6 +309,22 @@ public Dataset(String title, int[] dimSizes) throws DatasetException {
         } catch (IOException ioe) {
             throw new DatasetException("Can't create dataset " + ioe.getMessage());
         }
+        if (addFile) {
+            addFile(title);
+        }
```
JannalieT, line 314: The datasets add themselves to the Project...since the memory file constructor is used in processing when it is not wanted to add the dataset to the project, I added a flag so the memory file dataset can be added to the projects when needed

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/rs2d/RS2DData.java
```diff
@@ -149,6 +149,7 @@ public Dataset toDataset(String datasetName) throws IOException {
                 false, ByteOrder.BIG_ENDIAN, 0);
         dataset.newHeader();
         for (int i = 0; i < nDim; i++) {
+            resetSW(i);
```
JannalieT, line 152: Tested loading the input directory processed datasets, and JRES had an incorrect SW, this fixed it, and didn't break loading any of the other datasets in the test set. There is a todo in this file for loading datasets which may be why the SW wasn't correct but wasn't sure the reasoning behind the todo so didn't make any changes for it.
https://github.com/nanalysis/nmrfx/blob/2b397999fa056f8977e215a835618e181fb2c6f7/nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/rs2d/RS2DData.java#L264

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/jcamp/JCAMPData.java
```diff
@@ -840,4 +844,88 @@ public static boolean findData(StringBuilder bpath) {
...
+            complex.setFreqDomain(true);
+            dataset.writeVector(complex, index, 0);
+        }
+        // Set the header information in the dataset
+        for (int i = 0; i < dataset.getNDim(); i++) {
+            dataset.setValues(i, getValues(i));
+            // The first dimension is set when the vectors are set.
+            if (i > 0) {
+                dataset.setComplex(i, false);
+            }
```
JannalieT, line 913: This method is based off of the RS2D.toDatasets method, there the complex mode was set using i==0, so the first dimensions was always complex and the other dimensions were always false. Had some difficulty trying to reach a solution where both the T1 and JRES datasets loaded properly (tried using i==0 and using isComplex(i)) so ended up using whatever writeVector on line 905 set complex for the first dimension and setting all other dimensions to always false. Not sure if this is acceptable?

JannalieT:
tested with thais , loaded 1D processed dataset from the 60MHz okay (processed 2D datasets were never implemented on the 60), tested with spinlab exported JCAMP files, the datasets load at a higher level than expected but that can be adjusted using the Higher/lower buttons or ctrl + scroll. and then they look fine. Thais was able to use the dropdown menu to switch between the 2D and 1D datasets, (stacked datasets are not yet implemented, Jira issue for that tbd @brucejohnson)

brucejohnson (approved):

PR #143, opened 2022-07-13
NMR-5311 processing slider incorrect initial values

JannalieT:


darbymcgraw (approved):
Reproduced the bug on master and copied the same steps on your branch. Bug is fixed!

PR #60, opened 2022-01-21
Nmr 4685 gridpane

brucejohnson:

PR #144, opened 2022-07-15
Nmr 5313 export png exports blank image

JannalieT:
Exporting a blank png seems to have started when PolyChart was switched to extending a Node (https://github.com/nanalysis/nmrfx/pull/60). It seems like javafx Node.snapshot only takes a snapshot of whatever is both stored in a nodes children and is visible (and in the children's children etc).  The chartGroup object contains PolyChart objects as its children but the PolyChart objects don't have any children, therefore only a blank image was produced. The canvases can't be set to be the children of the PolyChart because nodes can only have one parent and there is more than one PolyChart but only one of each canvas. The chartPane object has the canvases as its children so using it for the snapshot will save the content, but the cursor lines must be disabled before snapshot and reenabled afterwards.

### DIFF
--- a/nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/utils/FileExtensionFilterType.java
+++ b/nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/utils/FileExtensionFilterType.java
@@ -1,0 +1,33 @@
+package org.nmrfx.processor.gui.utils;
+
+import javafx.stage.FileChooser.ExtensionFilter;
+import org.nmrfx.processor.datasets.vendor.rs2d.RS2DData;
+
+import java.util.List;
+
+/**
+ * Enum to keep file chooser extension descriptions consistent for each set of extension filter.
+ */
+public enum FileExtensionFilterType {
+    ALL_FILES("All Files", "*.*"),
+    NMR_FID("NMR Fid","fid", "ser", "*.nv", "*.dx", "*.jdx", "*.jdf", "*.dat"),
+    NMR_DATASET("NMR Dataset", "*.nv", "*.ucsf", "*.dx", "*.jdx", "1r", "2rr", "3rrr", "4rrrr", RS2DData.DATA_FILE_NAME),
+    PDF("PDF", "*.pdf"),
+    PNG("PNG", "*.png"),
+    SVG("SVG", "*.svg");
+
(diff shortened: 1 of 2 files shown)


nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/utils/FileExtensionFilterType.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.List;
+
+/**
+ * Enum to keep file chooser extension descriptions consistent for each set of extension filter.
+ */
+public enum FileExtensionFilterType {
+    ALL_FILES("All Files", "*.*"),
+    NMR_FID("NMR Fid","fid", "ser", "*.nv", "*.dx", "*.jdx", "*.jdf", "*.dat"),
+    NMR_DATASET("NMR Dataset", "*.nv", "*.ucsf", "*.dx", "*.jdx", "1r", "2rr", "3rrr", "4rrrr", RS2DData.DATA_FILE_NAME),
```
darbymcgraw, line 14: do we want .dat for NMR_DATASET too?

darbymcgraw (approved):
May want to add .dat to the NMR_DATASET filter type? Although you may have excluded it as it's not relevant there but I thought it was. Regardless, the exports have the file extensions now, and the .png displays an image!

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/utils/FileExtensionFilterType.java
JannalieT: RS2DData.DATA_FILE_NAME is a constant for `data.dat`
Both the RS2D FID files and the dataset files require the filename to be `data.dat` or they won't be loaded.  So the filter type for both NMR_FID and NMR_DATASET should probably be the same. Either `*.dat` or `data.dat`?  @brucejohnson or @vfiack could confirm which one to use?

brucejohnson: I'd use data.dat, as I don't think there are any other files that end .dat.

brucejohnson (dismissed):

brucejohnson (approved):

PR #146, opened 2022-07-19
NMR-5366 opening multiplet tool for 2D causes multiple errors

JannalieT:
Multiple messages/tools were a result of calling `getAnalyzer` on each button press without a null check. For example the initMultiplet resulted in 3 calls to getAnalyzer which resulted in three identical popup warnings. The regions tool had similar issues. Additionally only some multiplet buttons gave error popups, others resulted in NPE or ran on a previously loaded dataset as getAnalyzer was not called at the start of every method.
This pr:
- Updates analyzer when dataset is opened/active chart changed so it should always be up to date
- adds missing null checks to avoid NPE
- Disables the 1D menu option so region and multiplet tool cannot be accessed when a non 1D datataset is displayed *VISIBLE BEHAVIOUR CHANGE*
- Closes the tool if an incompatible dataset is loaded *VISIBLE BEHAVIOUR CHANGE*


JannalieT:
This fix is no longer needed with NMR-5169

PR #147, opened 2022-07-20
NMR-5310: set max width of bottom VBox to same maxWidth as processor …

JannalieT:
…tab pane, added a tooltip hover for text that was too long to fit in statusBar

### DIFF
--- a/nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/ProcessorController.java
+++ b/nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/ProcessorController.java
@@ -1114,6 +1114,9 @@ public class ProcessorController implements Initializable, ProgressUpdater {
         statusBar.setProgress(0.0);
 
         statusBar.getLeftItems().add(statusCircle);
+        Tooltip statusBarToolTip = new Tooltip();
+        statusBarToolTip.textProperty().bind(statusBar.textProperty());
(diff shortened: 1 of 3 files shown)


nmrfx-processor-gui/src/main/resources/fxml/ProcessorScene.fxml
```diff
@@ -148,7 +148,7 @@
         </TabPane>
     </center>
     <bottom>
-        <VBox>
+        <VBox maxWidth="400.0">
```
tbparis, line 151: Ok, that what limit the size.

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/ProcessorController.java
```diff
@@ -1114,6 +1114,9 @@ public void changed(ObservableValue<? extends Number> observableValue, Number nu
         statusBar.setProgress(0.0);
 
         statusBar.getLeftItems().add(statusCircle);
+        Tooltip statusBarToolTip = new Tooltip();
+        statusBarToolTip.textProperty().bind(statusBar.textProperty());
+        statusBar.setTooltip(statusBarToolTip);
```
tbparis, line 1119: And that's the tooltip to enable to see all the error.

tbparis (approved):
Tested and seems to works as expected (tested with a FID load from my Spinit data, process TDCOMB).

PR #151, opened 2022-07-27
NMR-5315: if processor is in use, dont load the scan table row

JannalieT:
~~If processor was busy, logged a message instead of trying to load the row.~~
Disabled scanner table and buttons if the processor is in use.

### DIFF
--- a/nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/tools/ScanTable.java
+++ b/nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/tools/ScanTable.java
@@ -55,7 +55,6 @@ import org.nmrfx.processor.gui.PolyChart;
 import org.nmrfx.processor.gui.ProcessorController;
 import org.nmrfx.processor.gui.controls.FileTableItem;
 import org.nmrfx.processor.gui.spectra.DatasetAttributes;
-import org.nmrfx.processor.processing.Processor;
 import org.nmrfx.utils.GUIUtils;
 import org.python.util.PythonInterpreter;
 import org.slf4j.Logger;
@@ -254,9 +253,6 @@ public class ScanTable {
                 }
             }
             chart.refresh();
-
-        } else if (!Processor.getProcessor().isProcessorAvailable()) {
-            log.warn("Unable to load row. Processor unavailable.");
         } else {
             openSelectedListFile();
         }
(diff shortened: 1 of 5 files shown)


nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/tools/ScanTable.java
```diff
@@ -254,6 +255,8 @@ final protected void selectionChanged() {
             }
             chart.refresh();
 
+        } else if (!Processor.getProcessor().isProcessorAvailable()) {
+            log.warn("Unable to load row. Processor unavailable.");
```
JannalieT, line 259: not sure if I should also add a popup to the user here as well?

tbparis: I would vote for no popup.
But what happens on selection, do user see the selection works on the tableview? I cannot test yet (asked you on Teams for data to reproduce the bug).
I'm just wondering if preventing user from selecting sth during the processing wouldn't be needed too. I didn't check if it is easily feasible so it is just an idea.

JannalieT: Yes on selection the row should be loaded, I think it could also be possible to disable the table from selection while the processor is busy though, I would just need to add a listener

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/tools/ScannerTool.java
```diff
@@ -650,4 +663,18 @@ void showTRACTGUI() {
         tractGUI.showMCplot();
     }
 
+    /**
+     * Disable the options if the processor is busy processing.
+     * @param newStatus
+     */
+    public void processorAvailableStatusUpdated(Boolean newStatus) {
```
darbymcgraw, line 670: This would be better as `boolean` as I don't think there's ever a need for `newStatus` to be `null`. Also `setDisable()` takes in a boolean

darbymcgraw (changes requested):

darbymcgraw:
Tested and it works. Cannot reproduce the bug on this branch as the fid files cannot be clicked.

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/tools/ScannerTool.java
JannalieT: was a copy paste typo, fixed it here and the other two spots it should have been `boolean`

darbymcgraw (approved):

brucejohnson (approved):

PR #152, opened 2022-07-28
Nmr 5442 save deuterium relaxation parameters when exporting r 1 r 2

brucejohnson:


nmrfx-core/src/main/java/org/nmrfx/chemistry/relax/RelaxationData.java
```diff
@@ -25,21 +25,31 @@
...
 
+    public String getId() {
+        return id;
+    }
+
+
+    public ResonanceSource getResSource() {
+        return resSource;
+    }
+
```
JannalieT, line 44: This one seems like it might be redundant? What is the reason to have both `getResSource` and `getResonanceSource`?

nmrfx-core/src/main/java/org/nmrfx/chemistry/relax/RelaxationData.java
```diff
@@ -25,21 +25,31 @@
...
 /**
  *
  * @author mbeckwith
  */
 public class RelaxationData implements RelaxationValues {
 
+    public String getId() {
+        return id;
+    }
+
```
JannalieT, line 39: I think moving these methods lower down might fix a lot of the sonar lint issues, variables should be moved above methods

nmrfx-core/src/main/java/org/nmrfx/chemistry/relax/RelaxationData.java
```diff
@@ -173,48 +178,67 @@ public Map<String, String> getExtras() {
...
+            }
+            Double value = data != null ? data.getValue() : null;
+            Double error = data != null ? data.getError() : null;
+            RelaxationValues.appendValueError(sBuilder, value, error,"%.3f");
+        }
+        return sBuilder.toString();
+    }
+
     public static void writeToFile(File file) throws IOException {
         MoleculeBase moleculeBase = MoleculeFactory.getActive();
```
JannalieT, line 219: add comment here about why its unused so method doesn't get removed accidentally in future

nmrfx-core/src/main/java/org/nmrfx/chemistry/relax/RelaxationData.java
brucejohnson: I think this (and the previous, getResSource) issue is from using IntelliJ to auto add methods.  It's not very good at putting them in the right place, and I'm not very good at checking where it put them.  I'll move them.

JannalieT (approved):

PR #153, opened 2022-07-28
Nmr 5169 implement multiplet tool etc as contextual popover

brucejohnson:


JannalieT (commented):
Tested with 1 and 2 charts showing:

Problems:

1. ~~Opened the wrong kind of molecule file and there was a parse exception. Could no longer open the molecule prompt, to select another molecule.~~
2. ~~Clicking split twice: first time splits the multiplet and then a No Value Found exception~~
3. ~~If the report is visible, and you switch to a 2D spectrum, the report stays visible,~~
4. ~~If you have two charts open, and you select a report to be visible on the first report, and then select the second chart and click the analyze drop down, the show report is shown as selected and you have to unselect and then reselect to view the report~~
5. ~~Clicking Integrate then Analyze then multiplets are calculated and drawn. Hitting Integrate and then Analyze a second time will result in a null pointer exception~~
6. ~~The Integrate Clear analysis cancel button doesn't stop Integrate (the cancel button the popup from clicking Clear, does stop clear from happening)~~
7. ~~Show Report when no chart loaded has two chart must have a 1D dataset popups~~
8. ~~Integrate button has a NullPointerException when you don't have a chart loaded~~
9. ~~If two charts are loaded, clicking Integrate on the first one will proceed, selecting the second chart and then clicking Integrate will do the would you like to clear prompt even though no analysis is done on the second chart (clicking yes does not affect the first chart that has analysis)~~
10. There seems to be some inconsistent issues with the crosshairs/selector on the mouse, sometimes it’s the crosshairs when its supposed to be the selector and vice versa - Jira issue made NMR-5459
11. ~~Analyze a dataset, load a different dataset, reload the first dataset, only the regions are displayed, then click analyze and everything is analyzed to singlets wherease before there were also multiplets~~

Suggestions

1. Multiplet popup always on top, on all applications is it possible to only be on top when the fxmlcontroller is on top
2. ~~Box title is always Info, regardless whether displaying multiplet info or region info, it might be helpful to update the title?~~
3. ~~Clearing the analysis/regions does not get rid of the popped out menu, so may want to atleast clear N: textbox~~
4. Changing the integral, causes the console window to popup, not really a bug just a bit annoying

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/tools/MultipletTool.java
```diff
@@ -805,65 +400,115 @@ void refresh() {
...
+                log.warn(ex.getMessage(), ex);
+            }
         }
     }
 
+
     public void addRegion(DatasetRegion region) {
-        getAnalyzer();
+        Analyzer analyzer = getAnalyzer();
+        System.out.println("add region " + analyzer);
```
JannalieT, line 511: Change to a log statement or remove
```suggestion
        log.info("add region " + analyzer);
```

JannalieT (approved):

PR #154, opened 2022-08-02
Fix inconsistent cursor modes

brucejohnson:


JannalieT (approved):
Tested this switching back and forth between the different cursors, the correct cursor was always showing. Only one note is that the cursor dropdown has changed size slightly
image

JannalieT (changes requested):
These sonar cloud issues should be fixed before merging.

JannalieT (approved):

PR #155, opened 2022-08-03
Fix extract/transfer/merge

brucejohnson:


JannalieT (approved):
tested extracting a peak from a multiplet and merging it back, transfering one peak to another multiplet, merging two multiplets and then splitting again

PR #156, opened 2022-08-04
NMR-5272: use the fpath to determine if BrukerData is an FID instead …

JannalieT:
…of isFrequencyDim

Unable to load any ser files as FID, switched to checking to see if the filenames exist at the filepath


brucejohnson:
I wonder if you should check if the BrukerData is an FID when it is opened and set a flag.  The current fix requires going to the file system to check for file existence every time isFID is called.

JannalieT:
> I wonder if you should check if the BrukerData is an FID when it is opened and set a flag. The current fix requires going to the file system to check for file existence every time isFID is called.

Set the flag in the constructor instead

darbymcgraw (approved):
Code looks good and tested by opening a Bruker FID file and the flag is set and the data is loaded.

PR #157, opened 2022-08-04
Nmr 5448 halting processing can cause exception on next load and process

JannalieT:
There were two separate issues:
1. On halt, the dataset was not always being closed, and trying to process a new FID would not create a new dataset since there was one already open. (stacktrace in HMBC-COSY.txt)
2. Loading files with different variable names for sw/sf resulted in an exception (stacktrace in hnconus-1H1d.txt)


darbymcgraw (approved):
Code looks good. Re-created the bug on master with HMBC & COSY. Did the same thing on this branch and the bug did not appear. Note that I was only able to recreate the bug if the process reached Process 2, and I could not recreate the bug by halting in Process 1.

brucejohnson (approved):

PR #158, opened 2022-08-10
Nmr 5420 tool buttons are inconsistent sizes

JannalieT:
Attempt to make tool buttons have a more consistent appearance and allow changing the icon/font sizes from one location.

Old tools
Old-AllToolsOpen

New Tools
New-AllToolsOpen


nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/MainApp.java
```diff
@@ -76,6 +78,11 @@ public class MainApp extends Application {
     protected static MainApp mainApp = null;
     static boolean isAnalyst = false;
     static Font defaultFont;
+    // Icon and font sizes for icon buttons
+    public static final String ICON_SIZE_STR = "16px";
+    public static final String ICON_FONT_SIZE_STR = "7pt";
+    // The default font size
+    public static final String REG_FONT_SIZE_STR = "9pt";
```
JannalieT, line 85: icon and font size for icon buttons can't be set in css files due to GlyphDudes.createIconButton method calls, the regular font size is also set here instead of in css so it may possible be changed programmatically in the future (request from @brucejohnson )

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/PeakNavigator.java
```diff
@@ -207,7 +208,20 @@ void initPeakNavigator(ToolBar toolBar, PeakNavigator parentNavigator, MenuButto
...
+            List<Node> navToolBarItems = this.navigatorToolBar.getItems().subList(1, this.navigatorToolBar.getItems().size());
+            Optional<Double> height = navToolBarItems.stream().map(node -> node.prefHeight(Region.USE_COMPUTED_SIZE)).max(Double::compare);
+            if (height.isPresent()) {
+                for (Node node : navToolBarItems) {
+                    if (node instanceof Control) {
+                        ((Control) node).setMaxHeight(height.get());
+                    }
+                }
+            }
+        });
```
darbymcgraw, line 224: Can you export these to a utility class? Since there's a lot of code repetition you could just pass the toolbar that you are looking to add this listener to.

darbymcgraw:
Windows: I think the sizing looks good. However, for the comparator, the rows are a different height (because the second item to compare doesn't have the close button). Also, for the comparator it looks strange that the arrows/buttons are in the middle of the screen. I think they should be on the left hand side.

Ubuntu: The sizing is off with the comparator in relation to the other toolbars. Also in the second set of arrows on the Path Tool are mismatched in size!
image

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/PeakNavigator.java
JannalieT: Moved this code to a method in GUIUtils

JannalieT:
> Windows: I think the sizing looks good. However, for the comparator, the rows are a different height (because the second item to compare doesn't have the close button). Also, for the comparator it looks strange that the arrows/buttons are in the middle of the screen. I think they should be on the left hand side.
> 
> Ubuntu: The sizing is off with the comparator in relation to the other toolbars. Also in the second set of arrows on the Path Tool are mismatched in size!

Made comparator rows same height and shifted arrows left, fixed sizing in ubuntu

darbymcgraw (approved):
Code looks good to me. GUI looks good in Windows as shown in Jannalie's comment. GUI matches throughout in linux
image

brucejohnson:
I think that for tools that don't have buttons with icons and text we should change the Close button to put the close icon to the left of "Close" ( ContentDisplay.LEFT) .  Otherwise all these tools have a height determined by the Close button and so have excess height over all the other buttons and menus.

JannalieT:
> I think that for tools that don't have buttons with icons and text we should change the Close button to put the close icon to the left of "Close" ( ContentDisplay.LEFT) . Otherwise all these tools have a height determined by the Close button and so have excess height over all the other buttons and menus.

Have switched the close icon to the left and also removed the icons from the peak slider
image

brucejohnson (approved):

PR #159, opened 2022-08-11
NMR-4506: store svg clipboard content as ByteBuffer

JannalieT:
Can copy and paste into desktop app Word (Office 365) on windows. (Right click after pasting to convert to shape and components can be moved around)

Can copy and paste into Inkscape (tested on Linux)

Note: Copy and paste does not work with online versions of Word (Office 365) which seems to have more limited copy and paste options: (https://support.microsoft.com/en-us/office/copy-and-paste-in-office-for-the-web-682704da-8360-464c-9a26-ff44abf4c4fe#__word_web_app)

### DIFF
--- a/nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/spectra/SpectrumMenuActions.java
+++ b/nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/spectra/SpectrumMenuActions.java
@@ -59,7 +59,7 @@ public class SpectrumMenuActions extends MenuActions {
         arrangeMenu.getItems().addAll(createGridItem, horizItem, vertItem, gridItem, overlayItem, minimizeItem, normalizeItem);
         MenuItem favoritesMenuItem = new MenuItem("Favorites");
         favoritesMenuItem.setOnAction(e -> showFavorites());
-        MenuItem copyItem = new MenuItem("Copy Spectrum as SVG Text");
+        MenuItem copyItem = new MenuItem("Copy Spectrum as SVG");
         copyItem.setOnAction(e -> FXMLController.getActiveController().copySVGAction(e));
         menu.getItems().addAll(newMenuItem, deleteItem, arrangeMenu, favoritesMenuItem, syncMenuItem, copyItem);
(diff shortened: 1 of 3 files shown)


nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/spectra/SpectrumMenuActions.java
```diff
@@ -59,7 +59,7 @@ public void basic() {
         arrangeMenu.getItems().addAll(createGridItem, horizItem, vertItem, gridItem, overlayItem, minimizeItem, normalizeItem);
         MenuItem favoritesMenuItem = new MenuItem("Favorites");
         favoritesMenuItem.setOnAction(e -> showFavorites());
-        MenuItem copyItem = new MenuItem("Copy Spectrum as SVG Text");
+        MenuItem copyItem = new MenuItem("Copy Spectrum as SVG");
```
JannalieT, line 62: removed `Text` as no longer copying as text, but as svg

darbymcgraw (dismissed):
Code looks good. Re-created the bug on the master branch and followed the steps on this branch. The image is now displayed when copying and pasting!

brucejohnson:
On Mac's I use Affinity Designer (cheaper, but very good, alternative to Adobe Illustrator.  It can paste in SVG in text format from the clipboard, but doesn't (I don't know why) accept it as a ByteBuffer.  I wonder if we should put two copies on the clipboard, in byte buffer and text formats. I don't know how widely this (paste to Affinity Designer), and one can work around it by saving to an sag file, but I at least use it fairly often.

JannalieT:
> On Mac's I use Affinity Designer (cheaper, but very good, alternative to Adobe Illustrator. It can paste in SVG in text format from the clipboard, but doesn't (I don't know why) accept it as a ByteBuffer. I wonder if we should put two copies on the clipboard, in byte buffer and text formats. I don't know how widely this (paste to Affinity Designer), and one can work around it by saving to an sag file, but I at least use it fairly often.

Added back the plain text svg on the clipboard. In word, ctrl + v, will paste the spectrum, and rightclick paste will paste the text.

brucejohnson (approved):
Now works again on Mac with Affinity Designer

PR #160, opened 2022-08-23
Nmr 5244 use spinlab dataset library

vfiack:


nmrfx-bom/pom.xml
```diff
@@ -19,6 +19,12 @@
 
     <dependencyManagement>
         <dependencies>
+            <dependency>
+                <groupId>com.nanalysis</groupId>
+                <artifactId>spinlab-dataset</artifactId>
+                <version>1.1.0</version>
```
vfiack, line 25: Repository for this project is here: https://github.com/nanalysis/spinlab-dataset

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/rs2d/RS2DData.java
```diff
@@ -235,29 +240,27 @@ private void openParFile(String parpath, boolean processed) throws IOException {
...
                 seriesDocument = readDocument(seriesPath);
             }
             groupDelay = readGroupDelay();
-            obsNucleus = header.getString(OBSERVED_NUCLEUS);
+            obsNucleus = header.get(OBSERVED_NUCLEUS).stringValue();
             Double obsFreq = null;
-            Double obsSW = header.getDouble(SPECTRAL_WIDTH);
-            tempK = header.getDouble(SAMPLE_TEMPERATURE);
-            String baseSFName = "";
+            double obsSW = header.<NumberValue>get(SPECTRAL_WIDTH).getValueAsHertz(header);
```
vfiack, line 251: This is the reason for the change: Spinlab data files can store SW as Hz or Ppm, reading the value directly from the xml is error-prone.

vfiack:
Build failing because dependency compiled for java17.

JannalieT (commented):
I tried loading a cosy from the data used for testing the jcamp data, and the axis labeling seems a bit off. For the processed spinlab datasets, it looks like values are off by a factor of 1e6 and the values for the unprocessed spinlabl datasets loaded and then processed, do not match the jcamp data at all.

cosy-spinlab

vfiack:
Interesting. I'll have to check with the current version to see if I broke something or if the import was already flawed.

JannalieT (approved):
retested with COSY and a JRES, looks good

PR #163, opened 2022-08-26
Nmr 5542 dataset link files nvlnk created when not needed

brucejohnson:


JannalieT (commented):
Tested saving a project and went to project files/datasets and all the dataset files were saved as .nv. (did not test with .ucsf)

JannalieT (approved):

PR #164, opened 2022-08-26
Nmr 5450 number format exceptions in ref manager

JannalieT:


nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/ChartProcessor.java
```diff
@@ -1110,14 +1110,17 @@ public String getScriptCmds(int nDim, String indent, boolean includeRun) {
...
+                            if (dimNum >= nDim) {
+                                break;
+                            }
+                            mapToDataset[dimNum] = -1;
+                        } catch (NumberFormatException nFE) {
+                            log.warn("Unable to parse dimension number.", nFE);
                         }
-                        mapToDataset[dimNum] = -1;
-                    } catch (NumberFormatException nFE) {
-                        log.warn("Unable to parse dimension number.", nFE);
```
JannalieT, line 1120: unsure if this change is right

brucejohnson (approved):

PR #165, opened 2022-08-31
NMR-5547: added new entrypoint for plugins MENU_FILE, and added new F…

JannalieT:
…ileExtensionType for txt files

Related to: https://github.com/nanalysis/nmrfx-private/pull/17

### DIFF
--- a/nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/FileMenuActions.java
+++ b/nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/FileMenuActions.java
@@ -49,8 +49,6 @@ public class FileMenuActions extends MenuActions {
         menu.getItems().addAll(openMenuItem, openDatasetMenuItem,
                 recentFIDMenuItem, recentDatasetMenuItem, datasetBrowserMenuItem,
                 graphicsMenu);
-
-        PluginLoader.getInstance().registerPluginsOnEntryPoint(EntryPoint.MENU_FILE, menu);
     }
 
     public void addAdvancedMenuItem() {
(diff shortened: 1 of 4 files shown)


nmrfx-plugin-api/src/main/java/org/nmrfx/plugin/api/EntryPoint.java
```diff
@@ -22,5 +22,5 @@
  * The obvious ones would be menus and scenes.
  */
 public enum EntryPoint {
-    STARTUP, MENU_PLUGINS, STATUS_BAR_TOOLS
+    STARTUP, MENU_PLUGINS, STATUS_BAR_TOOLS, MENU_FILE
```
vfiack, line 25: small detail but I'd prefer if we kept the similar entrypoints closer together.

something like that:
```suggestion
    STARTUP, MENU_FILE, MENU_PLUGINS, STATUS_BAR_TOOLS
```

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/FileMenuActions.java
```diff
@@ -47,6 +49,8 @@ public void basic() {
         menu.getItems().addAll(openMenuItem, openDatasetMenuItem,
                 recentFIDMenuItem, recentDatasetMenuItem, datasetBrowserMenuItem,
                 graphicsMenu);
+
+        PluginLoader.getInstance().registerPluginsOnEntryPoint(EntryPoint.MENU_FILE, menu);
```
vfiack, line 53: The other option was in makeMenuBar() in AnalystApp, but this works too.

vfiack (dismissed):

nmrfx-plugin-api/src/main/java/org/nmrfx/plugin/api/EntryPoint.java
JannalieT: made this change

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/FileMenuActions.java
JannalieT: moved to makeMenuBar

vfiack (approved):

PR #166, opened 2022-09-02
NMR-4505: add CopyMoleculeEventHandler to copy molecule to canvas if …

JannalieT:
…ctrl + v is pressed on canvas and to remove molecule from canvas if molecule is on canvas

This should work with a molecule copied straight from chem draw using ctrl + c (MDCLT format), or rightclick  -> copy (MDCLT format), or rightclick -> Molecule -> Copy as -> mol file (plain text). Also copy and pasting any mol file text from anywhere as long as it has the correct format and is in plain text, should paste a molecule


nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/AnalystApp.java
```diff
@@ -138,7 +134,7 @@ public void start(Stage stage) throws Exception {
         MainApp.setAnalyst();
         mainApp = this;
         analystApp = this;
-        FXMLController controller = FXMLController.create(stage);
+        FXMLController.create(stage).addCanvasKeyEventHandler(new CopyMoleculeEventHandler());
```
JannalieT, line 137: I'm not sure if this is the best setup, the CopyMoleculeEventHandler has to be added to the FXMLController each time a new one is created, but couldn't put this code in the `nmrfx-processor-gui` module because the canvas molecule code is in the `nmrfx-analyst-gui` module (would be circular dependencies), Should we move some of the molecule code to `nmrfx-processor-gui`? Or leave it where it is and set the event handler like this?, ...or other suggestions?

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/molecule/CopyMoleculeEventHandler.java
```diff
@@ -0,0 +1,93 @@
...
+            String filename = molContents.split("\n")[0];
+            try {
+                SDFile.read(filename, molContents);
+            } catch (MoleculeIOException e) {
+                log.error("Unable to read molecule file.");
+                return;
+            }
+            MoleculeUtils.addActiveMoleculeToCanvas();
+        } else if (event.getCode() == KeyCode.DELETE) {
+            MoleculeUtils.removeMoleculeFromCanvas();
```
JannalieT, line 53: Added a delete option in case someone changed their mind about the molecule they pasted, and there is no remove molecule option in the menu bar if the molecule is pasted into a 2D dataset or into an FID.

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/AnalystApp.java
brucejohnson: At some point I'd like to reevaluate the whole separation into nmrfx-processor-gui and nmrfx-analyat-gui.  There was a rational early on, where there were really two programs, NMRFx Processor that did just processing, and NMRFx Analyst that added lots of other features for analysis.  But we've pretty much gotten rid of NMRFx Processor" as a standalone GUI program.  But, I think we should make the addCanvasKey have a more generic function (than just molecules) and possibly register DataFormats from analyst.  So we'd register a DataFormat type for molecules from analyst code.  Needs a little more thought to see if this works.  It would also break your parsing of text as molecule.  But for that we may need a parser that can decide what the text represents (could be a molecule as done here, but could be other things like the text description of a spectrum).

JannalieT: Created a more general CanvasKeyEventHandler which can call different DataFormat handlers. I made a separate plain text handler so the pasting of plain text molecules still works, and we can add more discerning parsing there in the future if we need to paste something else in plain text.

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/events/MoleculeDataFormatHandler.java
```diff
@@ -0,0 +1,88 @@
...
+    @Override
+    public boolean handleDelete() {
+        PolyChart activeChart = FXMLController.getActiveController().getActiveChart();
+        Optional<CanvasAnnotation> anno = activeChart.hitAnnotation(activeChart.getMouseX(), activeChart.getMouseY());
+        if (anno.isPresent() && anno.get() instanceof CanvasMolecule) {
+            MoleculeUtils.removeMoleculeFromCanvas();
+            return true;
+        }
+        return false;
+    }
```
JannalieT, line 87: Will delete molecule if the user presses delete while their mouse is hovered over the molecule.

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/events/MoleculeDataFormatHandler.java
```diff
@@ -0,0 +1,88 @@
...
+        if (activeChart.getDataset() == null) {
+            return;
+        }
+        String molContents = parseMdlctBuffer((ByteBuffer)  o);
+        // Not a mol file without multiple lines
+        if (molContents.indexOf('\n') < 0) {
+            return;
+        }
+        // Use the first line of the string as the filename
+        String filename = molContents.split("\n")[0];
```
JannalieT, line 42: this may result in empty string filenames -> blank molecule names, not sure how to name the molecule if the title string isn't set otherwise..

JannalieT: discussed with @brucejohnson and added a prompt for a molecule name if there was no text on the title line of the pasted molecule file

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/AnalystApp.java
brucejohnson: There is an existing KeyHandler class in org.nmrfx.processor.gui.spectra.  Seems like we should handle the Paste events there rather than adding a new CanvasKeyEventHandler.
Also, instead of calling addCanvasDataFormatHandler on each created instance of FXMLController couldn't we just make that a static method add to static map in FXMLController?

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/events/MoleculeDataFormatHandler.java
brucejohnson: Similarly (to above) comment about KeyHandler, the Delete key is already handled in the KeyHandler class. We should use that code to call the molecule deletion

brucejohnson (approved):

PR #168, opened 2022-09-07
Remove punctuations from folder name

RemiBajart1:
(copied from spinlab)
if branch name has a punctuation sign (single apostrophe : ' in my case), build fails.


JannalieT (approved):
Is this a change that needs to be made to all github actions that are deleting from sharepoint? If so, I think there would be another place to change in the private nmrfx repository.

PR #169, opened 2022-09-07
NMR-5452: add delete button to integral tool and formatted to look mo…

JannalieT:
…re like multiplet tool, clear DatasetAttribute active region when setting the dataset


nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/spectra/DatasetAttributes.java
```diff
@@ -765,6 +765,7 @@ public void setDataset(DatasetBase aFile) {
         iBlkPut = new int[theFile.getNDim()];
         iWDim = new int[theFile.getNDim()];
         title = aFile.getTitle();
+        setActiveRegion(null);
```
JannalieT, line 768: If an active region was selected, and then another 1D dataset was loaded, the green active region bars were still visible at the old region

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/tools/IntegralTool.java
```diff
@@ -121,4 +144,17 @@ public void splitRegion() {
         }
 
     }
+
+    public void deleteRegion() {
```
JannalieT, line 148: The integral at the top is deleted, the multiplet is still visible though, this seems to be what the old RegionTool was doing as well...

JannalieT: now delete any peaks within the deleted region

brucejohnson (approved):

PR #170, opened 2022-09-08
Nmr 5584 remove module info

vfiack:
* remove module-info
* remove obsolete launchers (start using NMRAnalystApp class from IDE)
* replace ejml-fat jar with "normal" dependency


JannalieT (approved):
Tested running from NMRAnalystApp and nmrfx.bat (mvn package) as well as mvn install and they all work, I am not sure about the sonar lint Security Issue though.

vfiack:
> I am not sure about the sonar lint Security Issue though.

They say (rightfully) that allowing users to override the logging file may be an issue. 

That's more true for application servers than for local apps like this, here if you can do that, you already have access to the computer. Moreover, we don't call this with any user input, only static file path written by the developper => safe in our context.

PR #172, opened 2022-09-09
NMR-5580-molecules-displayed-on-canvas-are-upside-down

brucejohnson:


JannalieT (approved):
Tested by comparing with how 3 different molecules are displayed in chemdraw vs nmrfx canvas, the orientations match now.

PR #173, opened 2022-09-11
Nmr 5593 allow canvas objects molecules text with handles

brucejohnson:


nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/molecule/CanvasMolecule.java
```diff
@@ -333,52 +325,83 @@ String getHit() {
...
+            by1 = yPosType.move(startY1, dy, bounds[1], world[1]);
+        }
+        if (activeHandle == 2) {
+            bx2 = xPosType.move(startX2, dx, bounds[0], world[0]);
+            by2 = yPosType.move(startY2, dy, bounds[1], world[1]);
+        }
+        if (activeHandle == 3) {
+            bx1 = xPosType.move(startX1, dx, bounds[0], world[0]);
+            by2 = yPosType.move(startY2, dy, bounds[1], world[1]);
+        }
```
JannalieT, line 404: I think this should really be a series of it..else if, otherwise the condition gets checked for all even if the first condition passes

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/spectra/KeyBindings.java
```diff
@@ -128,7 +128,7 @@ public void keyPressed(KeyEvent keyEvent) {
                 case BACK_SPACE:
                     keyMonitor.complete();
                     keyEvent.consume();
-                    chart.deleteSelectedPeaks();
+                    chart.deleteSelectedItems();
```
JannalieT, line 131: I think the Delete key should be deleting the selected items as well as backspace, I would normally try deleting with "delete" before attempting to use "backspace" as a delete for anything that is not in a textfield

JannalieT (commented):
Tested and was able to select molecules and the analysis report, resize them and delete with backspace. I think you should also be able to delete with the delete key though

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/spectra/KeyBindings.java
JannalieT: merged DELETE and BACK_SPACE case

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/molecule/CanvasMolecule.java
```diff
@@ -333,52 +325,96 @@ String getHit() {
...
+            bx2 = xPosType.move(startX2, dx, bounds[0], world[0]);
+            by2 = yPosType.move(startY2, dy, bounds[1], world[1]);
+            bx2 = Math.max(bx1, bx2);
+            by2 = Math.max(by1, by2);
+        } else if (activeHandle == 3) { // bottom left
+            bx1 = xPosType.move(startX1, dx, bounds[0], world[0]);
+            by2 = yPosType.move(startY2, dy, bounds[1], world[1]);
+            bx1 = Math.min(bx1, bx2);
+            by2 = Math.max(by1, by2);
+        }
```
JannalieT, line 417: @brucejohnson when adjusting the molecules or the journal text, the handles no longer go past each other...but if you bring the handles right together, it can be really hard/impossible to reselect the object since there is no space, not sure if we should only allow the handles to get no less than x distance away from each other? and if so what should that distance be per position type?

brucejohnson: Yes, I think we should limit this.  A reasonable minimum might be twice the handle size

JannalieT: Added a limit of twice the handle size

JannalieT (approved):

PR #175, opened 2022-09-12
NMR-5198: add vm option to fix autocompletion binding to .bat files i…

JannalieT:
…n analyst-gui and processor-gui

Set JAVA_OPTS for the .bat files in the analyst-gui and processor-gui packages. This exception with the AutoCompletionBinding only seems to occur in windows, not linux so did not add anything to the bash scripts. Should be tested with mac though.

Related to this pr: https://github.com/nanalysis/nmrfx-private/pull/21


vfiack (approved):
strange that we need exports after removing modules.

JannalieT:
> strange that we need exports after removing modules.

@vfiack I followed the instructions for the non modular projects found in the controlsfx wiki: https://github.com/controlsfx/controlsfx/wiki/Using-ControlsFX-with-JDK-9-and-above

I am also not sure why the exception only happens when I run it in windows.

PR #177, opened 2022-09-14
NMR-4517 allow drag drop for molecules

JannalieT:
Molecule files can be dragged on to a canvas.
Highlighted molecule file text can be dragged on to a canvas.

Note: pr based off of NMR-4505, since modifying DataHandlers added in that class

### DIFF
--- a/nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/events/FilesDataFormatHandler.java
+++ b/nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/events/FilesDataFormatHandler.java
@@ -1,0 +1,67 @@
+package org.nmrfx.analyst.gui.events;
+
+import javafx.application.Platform;
+import org.nmrfx.analyst.gui.molecule.MoleculeUtils;
+import org.nmrfx.chemistry.io.MoleculeIOException;
+import org.nmrfx.chemistry.io.SDFile;
+import org.nmrfx.processor.datasets.vendor.NMRDataUtil;
+import org.nmrfx.processor.gui.PolyChart;
+import org.nmrfx.processor.gui.events.DataFormatEventHandler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.util.List;
+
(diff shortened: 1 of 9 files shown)


nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/events/FilesDataFormatHandler.java
```diff
@@ -0,0 +1,67 @@
...
+     * @param o List of Files to add.
+     * @param chart The chart to add the files to.
+     */
+    @Override
+    public boolean handlePaste(Object o, PolyChart chart) {
+        List<File> files = (List<File>) o;
+        if (files.isEmpty()) {
+            return false;
+        }
+        // Check the file type, if there are multiple files it is assumed they are all the same file type
```
JannalieT, line 36: This seems to be the previous behaviour for datasets (ex if the first file is a dataset it assumes all other files are also datasets) so did the same with molecules.

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/spectra/DragBindings.java
```diff
@@ -57,91 +72,45 @@ public void mouseDragDropped(final DragEvent e) {
...
+            isAccepted = SDFile.inMolFileFormat(contentString) || Dataset.getDataset(contentString.split("\n")[0]) != null;
         } else {
             e.consume();
         }
+        if (isAccepted) {
+            // fixme
+//                    chart.setStyle("-fx-border-color: green;"
+//                            + "-fx-border-width: 1;");
+            e.acceptTransferModes(TransferMode.COPY);
+        }
```
JannalieT, line 114: This fixme is from 2018, and was not related specifically to the jira issue, so did not fix, sonar lint will probably complain about this because I moved the code.

brucejohnson: Go ahead and remove this (chart.setStyle)

brucejohnson (dismissed):
Go ahead and remove the chart.setStyle commented lines

brucejohnson (approved):

PR #178, opened 2022-09-16
Nmr 5543 cannot select peaks after peak slider tool opened

JannalieT:
There were two sorts of issues:
1. When the peak slider tool was closed, the listeners weren't removed so setActivePeaks, which was drawing the peaks without selection, would end up be called after the selected peaks were drawn.
- fix -> remove listeners

The setActivePeaks method was drawing the peaks without selection so if the peak slider was open, selecting a multiplet arm would not show up as selected, even though it was selected and the popup multiplet tool would delete it.
- fix -> Dont call draw peaks in setActivePeaks method.

### DIFF
--- a/nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/tools/PeakSlider.java
+++ b/nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/tools/PeakSlider.java
@@ -10,17 +10,15 @@ import de.jensd.fx.glyphs.fontawesome.FontAwesomeIcon;
 import java.io.IOException;
 import java.util.*;
 import java.util.function.Consumer;
-import java.util.stream.Collectors;
 
+import javafx.beans.InvalidationListener;
 import javafx.event.ActionEvent;
 import javafx.event.Event;
-import javafx.scene.Node;
 import javafx.scene.control.*;
 import javafx.scene.input.MouseEvent;
 import javafx.scene.layout.HBox;
 import javafx.scene.layout.Pane;
 import javafx.scene.layout.Priority;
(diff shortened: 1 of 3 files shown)


nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/tools/PeakSlider.java
```diff
@@ -558,7 +590,6 @@ public void setActivePeaks(List<Peak> peaks) {
                                         });
                                     }
                                 }
-                                chart.drawPeakLists(true);
```
JannalieT, line 561: @brucejohnson double check this one especially. This seems to be working okay to me but I am not sure how the cluster matching peaks work, so I don't know if this breaks something.

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/tools/PeakSlider.java
```diff
@@ -173,26 +175,61 @@ public void initSlider(VBox vBox) {
...
+     * Add key bindings to newly added charts.
+     * @param change The change to the FXMLController charts list.
+     */
+    private void updateKeyBindings(ListChangeListener.Change<? extends PolyChart> change) {
+        if (change.next()) {
+            for (PolyChart chart: change.getAddedSubList()) {
+                addKeyBindingsToChart(chart);
+            }
+        }
+    }
```
JannalieT, line 209: added this listener so if a user added another chart while the peak slider was opened, the new chart would also have the peak slider key bindings set.

brucejohnson (approved):

PR #182, opened 2022-09-21
NMR-5454 integral norms wrong for 1D JCAMP spectrum

JannalieT:
The integrals are calculated twice, once in Analyzer (when the norm is set) and once before being drawn. The integrals calculated before being drawn were incorrect and were normalized with the correct norm, resulting in a lot of small normalized values.

The integrals calculated before being drawn were passing complex indices instead of raw indices to a method that called readPointRaw which mean the data was read from the incorrect spot in the file, and only half the needed length was read (second half of integral vector was all zeros).


nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/spectra/DatasetAttributes.java
```diff
@@ -1115,8 +1115,13 @@ public boolean getIntegralVec(Vec specVec, int iChunk, double ppm1, double ppm2,
...
         specVec.resize(dimSize, false);
-        ptC[0][0] = pt1;
-        ptC[0][1] = pt2;
+        if (theFile.getComplex(dim[0])) {
+            ptC[0][0] = pt1 * 2;
+            ptC[0][1] = (pt2 * 2) + 1;
+        } else {
+            ptC[0][0] = pt1;
+            ptC[0][1] = pt2;
+        }
```
JannalieT, line 1124: change complex indices to raw indices

nmrfx-analyst/src/main/java/org/nmrfx/analyst/peaks/Analyzer.java
```diff
@@ -139,6 +139,10 @@ public void calculateThreshold() {
         int size = dataset.getSizeTotal(0);
         Vec vec = new Vec(size);
 
+        if (dataset.getComplex(0)) {
+            vec.resize(size / 2, true);
+        }
+
```
JannalieT, line 145: For the complex datasets the vector ended up being the total size (ex. 16384) long instead of the complex size (ex. 8192) long which resulted in the second half of the vector being full of zeros. This resulted in one of the JCAMP datasets (INEPT_nd) having an exception when trying to integrate. It also meant that for complex datasets, the sDev value on line 160 will always be 0.  https://github.com/nanalysis/nmrfx/blob/36175bede8eaeb668aef62456697c557bdc545f1/nmrfx-analyst/src/main/java/org/nmrfx/analyst/peaks/Analyzer.java#L160

JannalieT:
@brucejohnson There may be more places we should check if Dataset.readVectorFromDatasetFile is getting passed the correct indices, since its used in 26 places, but I didn't want to do any more changes in case this isn't the right solution

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/peaks/PeakFitter.java
```diff
@@ -882,7 +892,19 @@ public double doFit(int i0, int i1, int[] rows, boolean doFit, boolean linearFit
...
+        int[][] ptRaw = new int[pt.length][2];
+        if (theFile.getComplex(pdim[0])) {
+            ptRaw[0][0] = pt[0][0] * 2;
+            ptRaw[0][1] = (pt[0][1] * 2) + 1;
+        } else {
+            ptRaw[0][0] = pt[0][0];
+            ptRaw[0][1] = pt[0][1];
+        }
+
+        theFile.readVectorFromDatasetFile(ptRaw, fitDim, fitVec);
```
JannalieT, line 907: When getting the vector for fitting the peaks, the raw indices also need to be used to retrieve the vector, but also need to keep the real/complex indices because the fitting function adjusts those indices further down in the function

brucejohnson (approved):

PR #184, opened 2022-09-26
Nmr 5449 negative integrals not displaying properly

JannalieT:

PR #187, opened 2022-09-29
Nmr 5707 jcamp spinlab datasets not saved in projects

JannalieT:
Spinlab datasets were saving as nvlnk files but weren't loading since only Bruker nvlnk files were set up.
JCAMP datasets were not saving since they didn't have a file set.

Also loaded the regions for the nvlnk files which were ignored previously, no way to store the norm with the nvlnk so I recalculated it

Note: based pr off https://github.com/nanalysis/nmrfx/pull/184 so that I can used the same norm calculations when loading the integrals.

### DIFF
--- a/nmrfx-core/src/main/java/org/nmrfx/datasets/DatasetFactory.java
+++ b/nmrfx-core/src/main/java/org/nmrfx/datasets/DatasetFactory.java
@@ -5,6 +5,9 @@
  */
 package org.nmrfx.datasets;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.io.IOException;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
@@ -15,13 +18,16 @@ import java.lang.reflect.Method;
  * @author brucejohnson
  */
 public class DatasetFactory {
+    private static final Logger log = LoggerFactory.getLogger(DatasetFactory.class);
+
+    private DatasetFactory() {}
 
     public static DatasetBase newDataset(String fullName, String name, boolean writable, boolean useCacheFile) throws IOException {
         DatasetBase dataset;
         try {
-            Class c = Class.forName("org.nmrfx.processor.datasets.Dataset");
-            Class[] parameterTypes = {String.class, String.class, boolean.class, boolean.class};
-            Constructor constructor = c.getDeclaredConstructor(parameterTypes);
+            Class<?> c = Class.forName("org.nmrfx.processor.datasets.Dataset");
+            var parameterTypes = new Class[]{String.class, String.class, boolean.class, boolean.class};
+            Constructor<?> constructor = c.getDeclaredConstructor(parameterTypes);
             dataset = (DatasetBase) constructor.newInstance(fullName, name, writable, useCacheFile);
         } catch (ClassNotFoundException | InstantiationException | IllegalAccessException | IllegalArgumentException | InvocationTargetException | NoSuchMethodException | SecurityException ex) {
             dataset = new DatasetBase(fullName, name, writable, useCacheFile);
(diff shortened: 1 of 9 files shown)


nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/jcamp/JCAMPData.java
```diff
@@ -888,7 +888,8 @@ public Dataset toDataset(String datasetName) throws IOException, DatasetExceptio
             datasetName = suggestName(fpath.toFile());
         }
         // Create a dataset in memory
-        Dataset dataset = new Dataset(datasetName + ".nv", dimSizes, true);
+        Dataset dataset = new Dataset(datasetName, dimSizes, true);
+        dataset.setFile(file);
```
JannalieT, line 892: setting the File even though the dataset is in memory mode so that when the project is saved it knows where to link to.

nmrfx-core/src/main/java/org/nmrfx/datasets/DatasetParameterFile.java
```diff
@@ -55,18 +55,10 @@ public String getParameterFileName() {
...
-        if (fileName.endsWith(".nv")) {
-            extLen = 3;
-        } else if (fileName.endsWith(".ucsf")) {
-            extLen = 5;
-        } else if (fileName.endsWith(".nvlnk")) {
-            extLen = 6;
-        }
-        parFileName = fileName.substring(0, len - extLen) + ".par";
-        return parFileName;
+        Pattern pattern = Pattern.compile(".(nv|ucsf|nvlnk)$");
```
tbparis, line 58: Shouldn't the pattern be ``\\.(nv|ucsf|nvlnk)$`` to detect the point?
If not a filename such as ``myfile_nvlnk`` become also ``myfile.par`` and ``myfilenvlnk`` become ``myfil.par``

nmrfx-core/src/main/java/org/nmrfx/datasets/DatasetFactory.java
```diff
@@ -15,31 +18,34 @@
...
+            var argTypes = new Class[]{String.class, String.class};
             Method m = c.getDeclaredMethod("newLinkDataset", argTypes);
             String[] args = {name, fullName};
-            System.out.println("new link " + fullName);
-            dataset = (DatasetBase) m.invoke(null, args);
+            log.debug("new link {}", fullName);
+            dataset = (DatasetBase) m.invoke(null, (Object[]) args);
         } catch (ClassNotFoundException | NoSuchMethodException | SecurityException | IllegalAccessException | IllegalArgumentException | InvocationTargetException ex) {
-            System.out.println("exc " + ex.getMessage());
+            log.warn(ex.getMessage(), ex);
```
tbparis, line 48: 👍

tbparis (approved):
Just few comment but tested on Spinlab dataset, on master I got the bug, on this branch not -> it works.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/Dataset.java
```diff
@@ -499,15 +501,29 @@ public static Dataset createDataset(String fullName, String title, int[] dimSize
...
         log.info("{}", nmrData);
-        BrukerData brukerData = (BrukerData) nmrData;
-        return brukerData.toDataset(name);
+        if (nmrData instanceof BrukerData brukerData) {
+            return brukerData.toDataset(name);
+        } else if (nmrData instanceof RS2DData rs2DData) {
+            return rs2DData.toDataset(name);
+        } else if (nmrData instanceof JCAMPData jcampData) {
+            return jcampData.toDataset(name);
+        }
```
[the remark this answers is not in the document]
JannalieT: NMRData is how the data is initially loaded and is used with the FID data, the Dataset is for processed data but the NMRData class is used to load it. Some of the NMRData extended classes have the toDataset functions, but some of them like NMRViewData just have a dataset object in them that is created automatically by reading the datafile.

nmrfx-core/src/main/java/org/nmrfx/datasets/DatasetParameterFile.java
JannalieT: Yes, I forgot to escape it, changed this one and the same error in ProjectBase.java

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/jcamp/JCAMPData.java
JannalieT: I removed it because when the project was loaded the datasetName passed into dataset contains the .dx, so the filename ends up being <filename text>.dx.nv and then it was not recognized as the correct dataset, I need to take a closer look at this because it seems a bit sketchy for that to have solved the problem, and this probably isn't the best solution

JannalieT: added the dataset file as an option in the constructor to make it match better with the other Dataset constructors

brucejohnson (approved):

PR #188, opened 2022-09-30
Nmr 5730 rs2 d jcamp datasets cant drag and drop

JannalieT:
JCAMP processed data, RS2D unprocessed and processed data, Bruker processed data and Jeol jdf files (only tested with the two from the test files) can now be drag and dropped

### DIFF
--- a/nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/events/FilesDataFormatHandler.java
+++ b/nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/events/FilesDataFormatHandler.java
@@ -57,18 +57,22 @@ public class FilesDataFormatHandler implements DataFormatEventHandler {
 
     /**
      * Checks if first file in list is dataset is FID or processed. If it is an FID, the first file is opened, if
-     * it is processed, all the files are opened in  append mode.
+     * it is processed, all the files are opened in dataset append mode.
      * @param files The files to open.
      * @param chart The chart to display the files in.
      */
     private void openDataFiles(List<File> files, PolyChart chart) {
         chart.setActiveChart();
-        boolean isFID;
+        boolean isFID = false;
         try {
             isFID = NMRDataUtil.getFID(files.get(0).getAbsolutePath()).isFID();
         } catch (IOException e) {
-            log.warn("Unable to open datafiles. {}", e.getMessage(), e);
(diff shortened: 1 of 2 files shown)


nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/NMRDataUtil.java
```diff
@@ -177,8 +177,12 @@ public static String isFIDDir(String fpath) {
             return bpath.toString();
         } else if (VarianData.findFID(bpath)) {
             return bpath.toString();
+        } else if (RS2DData.findFID(bpath)) {
+            return bpath.toString();
         } else if (JCAMPData.findFID(bpath)) {
             return bpath.toString();
+        } else if (JeolDelta.findFID(bpath)) {
+            return bpath.toString();
```
JannalieT, line 185: Tested with the Jeol jdf files from the testfids we have, and they also couldn't drag and drop so added this here. Didn't add anything for JEOL in the isDatasetFile method, as it looks we don't load Jeol files as datasets anywhere else.

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/events/FilesDataFormatHandler.java
```diff
@@ -49,19 +50,36 @@ public boolean handlePaste(Object o, PolyChart chart) {
...
+        boolean isFID = false;
+        try {
+            isFID = NMRDataUtil.getFID(files.get(0).getAbsolutePath()).isFID();
+        } catch (IOException e) {
+            // If the file can't be opened as an FID NMRData, check if it is a datafile, since processed Bruker files
+            // will not open with NMRDataUtil.getFID, otherwise return
+            if (NMRDataUtil.isDatasetFile(files.get(0).getAbsolutePath()) == null) {
+                log.warn("Unable to open datafiles. {}", e.getMessage(), e);
+                return;
+            }
```
JannalieT, line 75: processed Bruker files need this extra check as they won't load as NMRData with getFID.

@brucejohnson  Note: the drag and drop works for the 1r, 1i, and 2rr files only from the processed files shared. These are also the only 3 files that can be opened using "Open Dataset"

brucejohnson (approved):

PR #189, opened 2022-09-30
NMR-5729 projects generating too many windows

JannalieT:
The stages and controllers were not being removed from the stages/controllers lists since the close functionality was not being called. setOnCloseRequest is only triggered by external close events, programmatically closing will not trigger it.


brucejohnson:
I think we should keep stage.setOnCloseRequest instead of setOnHidden.  There's a bug in get stages from the scripting commands (nd.stages(), but if that was working it would be possible to get a stage and call hide (or close) on it to temporarily hide it, but then one couldn't call show(), as the reference to the stage will be lost.  To avoid the current problem I think closeAll (the only place, I think, where stages are currently closed with the intent that it be permanent) should itself call controller.close() and removeStage(stage).

brucejohnson (approved):

PR #195, opened 2022-10-21
Nmr 5803 improve 1 d projection display

JannalieT:
Dragging 1D dataset onto 2D dataset will resize the border and add as a projection to the matching axis.
Projections can be removed through the Extract Slice/ Projection menu.
Projections can be scaled using Attributes -> Slices -> Scale slider which will scale both projections together

Scaling still needs to be improved; changes made in: https://github.com/nanalysis/nmrfx/pull/196

This pr does not cover the case of what should happen if a user drags a 2D dataset onto a 1D dataset (Jira issue made: NMR-5731)

### DIFF
--- a/nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/ChartProperties.java
+++ b/nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/ChartProperties.java
@@ -37,6 +37,7 @@ import java.util.Map;
 public class ChartProperties {
     private static final Logger log = LoggerFactory.getLogger(ChartProperties.class);
     public static final int PROJECTION_BORDER_DEFAULT_SIZE = 150;
+    public static final int EMPTY_BORDER_DEFAULT_SIZE = 7;
 
     final private PolyChart polyChart;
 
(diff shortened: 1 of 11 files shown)


nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/ChartProperties.java
```diff
@@ -36,6 +36,8 @@
  */
 public class ChartProperties {
     private static final Logger log = LoggerFactory.getLogger(ChartProperties.class);
+    public static final int PROJECTION_BORDER_DEFAULT_SIZE = 150;
+    public static final int EMPTY_BORDER_DEFAULT_SIZE = 7;
```
JannalieT, line 40: 7 is the initial value for the border, but it is a bit strange since 7 is not one of the options in the menu size combobox in Attributes -> Style tab. So once you click on the combo box you can never get back to 7. Perhaps we should change it to a default value of 5?

brucejohnson: makes sense. I don't know how I came up with seven. Unless it just looked good

brucejohnson (dismissed):

brucejohnson (approved):

PR #196, opened 2022-10-24
Nmr 5831 improve projection scaling

JannalieT:
based off https://github.com/nanalysis/nmrfx/pull/195

### DIFF
--- a/nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/spectra/GestureBindings.java
+++ b/nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/spectra/GestureBindings.java
@@ -77,8 +77,8 @@ public class GestureBindings {
         if (event.isControlDown() && border == ChartBorder.NONE) {
             chart.scaleY(dy);
         } else if (border == ChartBorder.RIGHT || border == ChartBorder.TOP) {
-            // Divide by 100 to make it easier to scale, doesn't change as quickly
-            chart.updateProjectionScale(border, dy / 100);
+            // Divide by 1000 to make it easier to scale, doesn't change as quickly
+            chart.updateProjectionScale(border, dy / 1000);
             chart.refresh();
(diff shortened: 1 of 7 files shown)


brucejohnson:
If you use the scroll wheel in the 2D spectrum the projections scale in the opposite direction to what happens if you scroll in the projection

JannalieT:
> If you use the scroll wheel in the 2D spectrum the projections scale in the opposite direction to what happens if you scroll in the projection

Disabled the projection scrolling while scrolling in the main spectrum

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/spectra/GestureBindings.java
```diff
@@ -71,12 +71,16 @@ public void zoom(Event event) {
...
+        ChartBorder border = chart.hitBorder(x, y);
         double dx = event.getDeltaX();
         double dy = event.getDeltaY();
-        if (event.isControlDown()) {
+        if (event.isControlDown() && border == ChartBorder.NONE) {
             chart.scaleY(dy);
-        } else if ((border != 0) || event.isAltDown()) {
+        } else if (border == ChartBorder.RIGHT || border == ChartBorder.TOP) {
+            // Divide by 1000 to make it easier to scale, doesn't change as quickly
+            chart.updateProjectionScale(border, dy / 1000);
```
brucejohnson, line 81: Projections still scale opposite to what I would expect.  If you scroll a scroll wheel up in a 1D spectrum the spectrum gets higher, but if you do it in a projection the projection gets smaller.  Changing the sign of dy will reverse this.  Also, I find the scrolling too slow, maybe divide by 500 rather than 1000 :  chart.updateProjectionScale(border, -dy / 500)

JannalieT: scroll directions now consistent between projections in 1D, but scrolling direction is reversed in Windows to what I would expect. As discussed on teams this is due to natural vs traditional mouse scrolling. This should be fixed in NMR-4777

brucejohnson (approved):

PR #198, opened 2022-10-28
Nmr 5848 use nus processing as alternative to linear prediction

brucejohnson:


nmrfx-processor/src/main/java/org/nmrfx/processor/operations/GRINSOp.java
```diff
@@ -87,16 +87,88 @@ public GRINSOp(double noise, double scale, int zfFactor,
...
+    public Operation evalMatrix(MatrixType matrix) {
+        if (extendMode) {
+            return evalExtendMatrix(matrix);
+        } else {
+            return evalNUSMatrix(matrix);
+        }
+
+    }
+
+
```
JannalieT, line 132: extra space
```suggestion

```

nmrfx-processor/src/main/java/org/nmrfx/processor/operations/NESTANMR.java
```diff
@@ -131,19 +179,58 @@ public Operation eval(Vec vector) throws ProcessingException {
...
+            return evalExtendMatrix(matrix);
+        } else {
+            return evalNUSMatrix(matrix);
+        }
+
+    }
+
+    public static int getZfSize(double vecSize, int factor) {
+        return (int) (Math.pow(2, Math.ceil((Math.log(vecSize) / Math.log(2)) + factor)));
+    }
```
JannalieT, line 200: exact duplicate of static method of MatrixND.getZFSize, only difference is this one is public and the other one is default access...do we need both methods?

JannalieT (approved):

PR #199, opened 2022-10-31
Nmr 5888 forward backward linear prediction artifacts

brucejohnson:


JannalieT (dismissed):
tested with attached dataset, artifact at 8.54 ppm was gone in this branch

JannalieT (approved):

PR #200, opened 2022-11-01
Nmr 5715 add region table

JannalieT:


nmrfx-analyst/src/main/java/org/nmrfx/analyst/peaks/Multiplets.java
```diff
@@ -1132,20 +1130,14 @@ public static void splitRegionsByPeakCount(Set<DatasetRegion> regions, PeakList
...
                 limits[0][1] = region.getRegionEnd(0);
-                DatasetRegion newRegion = null;
                 List<Peak> peaks = locatePeaks(peakList, limits, dim);
                 if (peaks.size() > maxPeaks) {
-                    List<PeakDim> peakDims = new ArrayList<>();
-                    peaks.forEach((peak) -> {
-                        peakDims.add(peak.peakDims[0]);
-                    });
                     int nSplits = peaks.size() / maxPeaks;
-                    peakDims.sort(comparing(PeakDim::getChemShiftValue));
```
JannalieT, line 1148: Sonar lint wanted lambda to be shorter, doesn't seem like peakDims is used for anything so removed it. This entire method splitRegionsByPeakCount is unused though...

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/regions/RegionsTable.java
```diff
@@ -0,0 +1,234 @@
...
+
+        TableColumn<DatasetRegion, Double> integralCol = new TableColumn<>(INTEGRAL_COLUMN_NAME);
+        integralCol.setCellValueFactory(new PropertyValueFactory<>("integral"));
+        integralCol.setCellFactory(column -> new DoubleTableCell(NUMBER_DECIMAL_PLACES_INTEGRAL));
+        getColumns().add(integralCol);
+
+        TableColumn<DatasetRegion, Double> normalizedIntegralCol = new TableColumn<>(NORMALIZED_INTEGRAL_COLUMN_NAME);
+        normalizedIntegralCol.setCellValueFactory(param -> {
+            double norm = chart.getDataset().getNorm() / chart.getDataset().getScale();
+            Double normProp = norm != 0.0 ? param.getValue().getIntegral() / norm : null;
```
JannalieT, line 133: sonar flagged this as a bug, is it alright to use != on doubles when trying to avoid a divide by zero? Or should I check with precison...and to what precision should I check?

brucejohnson:
If clicking on a region in table is going to highlight that region (which seems a good idea), we should jump the display to that region (use     public void moveTo(Double[] positions)  in PolyChart).  I'd use that one (with position at center of region so the width of view doesn't change.  I'd also use isInView to check if its already visible, if so, don't move.  Example in SimFitMolController regionChanged method

I'd add a normalize button, so a user could select a row, and then normalize table to that value (with a popup to request value to normalize to, like the popup on integral on spectrum)

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/regions/RegionsTable.java
brucejohnson: Something like Math.abs(norm) > 1.0e-9 should silence sonar, and be reasonable

JannalieT:
> If clicking on a region in table is going to highlight that region (which seems a good idea), we should jump the display to that region (use public void moveTo(Double[] positions) in PolyChart). I'd use that one (with position at center of region so the width of view doesn't change. I'd also use isInView to check if its already visible, if so, don't move. Example in SimFitMolController regionChanged method
> 
> I'd add a normalize button, so a user could select a row, and then normalize table to that value (with a popup to request value to normalize to, like the popup on integral on spectrum)
> 
> Seems funny to have the Show Regions Table in the Peak menu. Maybe it should be in the View menu for now

Added these items:

- When highlighting region - Jump to display
- Move to View menu
- Add a txt extension filename (utility function)
- Add clear button to clear the table
- Add AutoIntegrate Button to redo the whole integration 

Did not add:
- Normalize button -> can edit integral norms in the table, editing a row will adjust the rest of the table
- Sensitivity Control -> Perhaps in future jira issue?

brucejohnson:
There are multiple places where Regions are added that should be labeled as Manual, not Fitted.  I might invert the logic and default regions to Manual, and only set them as Auto from places like autoSetRegions in the Analyzer.
Might be convenient to also allow showing region table from the "Integrate" split menu button

JannalieT:
> There are multiple places where Regions are added that should be labeled as Manual, not Fitted. I might invert the logic and default regions to Manual, and only set them as Auto from places like autoSetRegions in the Analyzer. Might be convenient to also allow showing region table from the "Integrate" split menu button

Changes:

- Set Manual as default, also changed 'Fitted' to 'Auto' in the table
- Update 'Auto' label to 'Manual' if user modifies bounds
- Added table button to integral tool
- Default normalization on loading regions file
- Fixed axis bounds error (affected region selection and or drag selecting past bounds of chart)

brucejohnson (approved):

brucejohnson (approved):

PR #201, opened 2022-11-02
NMR-5803 projections with JCAMP RS2D datasets

JannalieT:
projections with complex datasets had 2 issues, the empty real vec was being read instead of real values from complex vec, and only half the file was read due to real indices being passed instead of raw to readDatasetFromFile

Also fixed exception that occurs if a user switched from 2D view to 1D view while projections are displayed

### DIFF
--- a/nmrfx-core/src/main/java/org/nmrfx/datasets/DatasetBase.java
+++ b/nmrfx-core/src/main/java/org/nmrfx/datasets/DatasetBase.java
@@ -15,11 +15,11 @@ import java.util.*;
 public class DatasetBase {
     private static final Logger log = LoggerFactory.getLogger(DatasetBase.class);
 
-    public final static int NV_HEADER_SIZE = 2048;
-    public final static int UCSF_HEADER_SIZE = 180;
-    public final static int LABEL_MAX_BYTES = 16;
-    public final static int SOLVENT_MAX_BYTES = 24;
-    public final static String DATASET_PROJECTION_TAG = "_proj_";
+    public static final int NV_HEADER_SIZE = 2048;
+    public static final int UCSF_HEADER_SIZE = 180;
+    public static final int LABEL_MAX_BYTES = 16;
+    public static final int SOLVENT_MAX_BYTES = 24;
+    public static final String DATASET_PROJECTION_TAG = "_proj_";
     public DatasetLayout layout = null;
     /**
      *
--- a/nmrfx-core/src/main/java/org/nmrfx/datasets/DatasetBase.java
+++ b/nmrfx-core/src/main/java/org/nmrfx/datasets/DatasetBase.java
@@ -8,6 +8,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.RandomAccessFile;
 import java.nio.ByteOrder;
 import java.util.*;
 
@@ -18,6 +19,7 @@ public class DatasetBase {
     public final static int UCSF_HEADER_SIZE = 180;
     public final static int LABEL_MAX_BYTES = 16;
     public final static int SOLVENT_MAX_BYTES = 24;
+    public final static String DATASET_PROJECTION_TAG = "_proj_";
     public DatasetLayout layout = null;
     /**
      *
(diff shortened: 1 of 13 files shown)


nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/spectra/DatasetAttributes.java
```diff
@@ -928,7 +928,7 @@ public void getProjection(Dataset dataset, Vec specVec, int iDim) throws IOExcep
         ptC[0][0] = dataset.ppmToPoint(0, ppm0);
         ptC[0][1] = dataset.ppmToPoint(0, ppm1);
         specVec.resize(ptC[0][1] - ptC[0][0] + 1, dataset.getComplex_r(0));
-        dataset.readVectorFromDatasetFile(ptC, dimC, specVec);
+        dataset.readVectorFromDatasetFile(DatasetUtils.generateRawIndices(ptC, dataset.getComplex_r(0)), dimC, specVec);
```
JannalieT, line 931: Dataset has both getComplex and getComplex_r ... they are very similar, I'm not sure which one to use?

brucejohnson: Should be getComplex here (though should work with both when not processing)

brucejohnson:
I should've caught this with the earlier projection pull request, but is it necessary to have a separate projection scale value. Can't we just use the existing level value in data attributes. I was thinking about this because we need to be able to save the state of windows with projections. If we add a new projection scale, we need to worry about exporting that value into the yaml file. If we just use the existing level value the only thing we have to add to the yaml file is a flag indicating that it's a projection.

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/spectra/DatasetAttributes.java
JannalieT: What about the line above it (930) should that also be getComplex?

JannalieT: changed both to getComplex

brucejohnson:
Unrelated to JCAMP, but other projection thoughts:
Seems like autoScale should call updateProjectionScale, and adjustScale and scaleY should skip projections.  Currently they scale together, which I was thinking we were moving away from.  What do you think?  I made those changes in my own copy so I could play with it.  I can push them to the branch if you'd like.

Also, another issue.  If you click Pick with projections on it tries, and fails, to pick them.  We should in the peakPick code skip over DatasetAttributes that are projections.  Related, with this and above code I do a lot of "if (projection() != -1)" etc., might be more clear if we added isProjection()

nmrfx-core/src/main/java/org/nmrfx/math/VecBase.java
```diff
@@ -131,7 +131,7 @@ public VecBase(int size, boolean complex) {
     public VecBase(int size, boolean complex, PyType type) {
         super(type);
         this.isComplex = complex;
-        useApache = true;
+        useApache = complex;
```
JannalieT, line 134: Changed this value here because it was set to true even when complex is false which lead to an exception in vecMat.getBytes since it tried to access cvec in a non complex vector that only had rvec set. Not sure if this is the best way to solve this?

nmrfx-core/src/main/java/org/nmrfx/datasets/DatasetBase.java
```diff
@@ -18,6 +19,7 @@ public class DatasetBase {
     public final static int UCSF_HEADER_SIZE = 180;
     public final static int LABEL_MAX_BYTES = 16;
     public final static int SOLVENT_MAX_BYTES = 24;
+    public final static String DATASET_PROJECTION_TAG = "_proj_";
```
JannalieT, line 22: This probably isn't the best spot for it since Datasets don't track if they are projections, but since I needed to access it from nmrfx-core, I couldn't put it in DatasetAttributes

nmrfx-core/src/main/java/org/nmrfx/math/VecBase.java
JannalieT: Undid this change, see comment about cvec null check

nmrfx-core/src/main/java/org/nmrfx/math/VecBase.java
```diff
@@ -1036,7 +1036,7 @@ public byte[] getBytes() {
                     buffer[j++] = (byte) (intVal & 0xFF);
                 }
             }
-        } else if (useApache) {
+        } else if (useApache && cvec != null) {
```
JannalieT, line 1039: A lot of real Vectors had useApache set to true, but cvec was null, rvec was used since it was a real vector. Instead of changing useApache to false for real vectors because that could possible affect a lot of areas in the code, just added a null check here since this is the first time getBytes is actually being used, it may not have been tested before. Double check if this is okay

brucejohnson: I think the problem is that the whole  } else if (useApache) {  branch shouldn't be there.  If the Vec is not complex, we should not check useApache, and should always use data in rvec.

brucejohnson (approved):

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/MainApp.java
```diff
@@ -92,7 +92,7 @@ public static void closeAll() {
             chart.clearDataAndPeaks();
             chart.clearAnnotations();
         }
-        List<FXMLController> controllers = FXMLController.getControllers();
+        List<FXMLController> controllers = new ArrayList<>(FXMLController.getControllers());
         // Don't close the first controller that matches with the main stage, Note this first controller is not
         // necessarily the active controller
         for (int index = 1; index < controllers.size(); index++) {
```
JannalieT, line 98: controllers.size was decreasing each loop due to the close call in each loop, so not all the controllers were closed,

brucejohnson (approved):

PR #202, opened 2022-11-03
Nmr 5917 spinit data hz ppm

vfiack:


nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/rs2d/RS2DData.java
```diff
@@ -409,28 +410,113 @@ public String getFilePath() {
...
+     *
+     * @param name The parameter name
+     * @return the floating point representation of its value.
+     */
     @Override
     public Double getParDouble(String name) {
+        Number convertedNumber = getConvertedNumberValue(name);
+        if (convertedNumber != null) {
+            return convertedNumber.doubleValue();
+        }
```
vfiack, line 447: This was the original bug: python processing was changing sw from the parameter value based on its name, and this didn't include the ppm -> Hertz conversion, so the spectral width was badly interpreted.

JannalieT (approved):
I do not have any recent spinit data files to test with so can't test for the bug fix so only tested with data from JCAMP Export/input, and those files still load matching the spectrum images.

PR #203, opened 2022-11-03
NMR-5841 scanner tool not working with JCAMP RS2D

JannalieT:
The drawlist was being set based on the number of frequency dimensions. For T1 experiments with .nv datasets, this value is set as 1 (value set from the .par file), for other types of datasets this value is set to the number of dimensions, (since they don't have a par file) so a RS2D t1 has this value set as 2. Since we only differentiate between 2D and pseudo2D by guessing from the number of scans, it would be difficult to set the number of frequency dimensions correctly. Using the chart display type property seems to work for both the .nv files and the JCAMP and RS2D datasets.

### DIFF
--- a/nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/tools/ScanTable.java
+++ b/nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/tools/ScanTable.java
@@ -230,7 +230,7 @@ public class ScanTable {
                 curLvl.ifPresent(dataAttr::setLvl);
                 int nDim = dataAttr.nDim;
                 chart.full(nDim - 1);
-                if ((nDim - dataAttr.getDataset().getNFreqDims()) == 1){
+                if (chart.getDisDimProperty().get() == PolyChart.DISDIM.OneDX){
                     chart.setDrawlist(rows);
                 } else{
                     chart.clearDrawlist();
--- a/nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/tools/ScanTable.java
+++ b/nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/tools/ScanTable.java
@@ -79,8 +79,8 @@ public class ScanTable {
 
     ScannerTool scannerTool;
     TableView<FileTableItem> tableView;
-    TableFilter fileTableFilter;
-    TableFilter.Builder builder = null;
+    TableFilter<FileTableItem> fileTableFilter;
+    TableFilter.Builder<FileTableItem> builder = null;
     File scanDir = null;
 
     PopOver popOver = new PopOver();
@@ -91,10 +91,11 @@ public class ScanTable {
     Set<String> groupNames = new TreeSet<>();
     Map<String, Map<String, Integer>> groupMap = new HashMap<>();
     int groupSize = 1;
-    ListChangeListener filterItemListener = (ListChangeListener.Change c) -> {
+    ListChangeListener<FileTableItem> filterItemListener = c -> {
         getGroups();
         selectionChanged();
     };
+    ListChangeListener<Integer> selectionListener;
 
     static final List<String> standardHeaders = List.of("path", "sequence", "row", "etime", "ndim");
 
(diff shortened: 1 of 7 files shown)


brucejohnson:
I'm not sure if this will always work, because not only can you have pseudo-two dimensional files you can have pseudo-three-dimensional files and they should display in the scanner tool. If one can't figure out from the data set that the number of dimensions is greater than the number of frequency dimensions, then we could figure it out when we use the scanner tool. If you do load from data set, you can assume that it's an arrayed experiment the last dimension be the arrayed dimension. If the data set is two dimensional, then display it as curves. And if it's three-dimensional then display it as contours. If the scanner tool is populated by scanning a directory we can assume that if the individual files are one dimensional display them as curves, if they're two dimensional display them as contours. If we are loading a scan table then we check to see if there is more than one data set mentioned in the table. If there's only one data set, then we can assume that it is in an arrayed  experiment. In that case, we display as curves if it's a two dimensional experiment, and as contours if it's a three dimensional experiment. There is more than one experiment, then we assume that the individual experiments aren't arrayed, the array is the series of experiments.

What I'm not sure about is whether if one deduces the type of experiment (arrayed) from the above criteria, if one should set the number of frequency dimensions appropriately. That way all the display of that experiment in the session will automatically be correct without having to set the curves/contours control.

I think for spinit and NMRFX files we can figure out whether it's arrayed. I guess the problem is with JCAMP files.  MNova claims to be able to detect whether the experiments are arrayed, but I don't know what they check for or whether it is reliable

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/tools/ScanTable.java
```diff
@@ -561,9 +562,18 @@ public void loadFromDataset() {
...
+            controller.openDataset(dataset.getFile(), false, false);
+            DatasetBase datasetReloaded = chart.getDataset();
             List<Integer> rows = new ArrayList<>();
             rows.add(0);
+            // Load from Dataset assumes an arrayed dataset
+            datasetReloaded.setNFreqDims(datasetReloaded.getNDim() - 1);
+            if (datasetReloaded.getNDim() > 2) {
+                chart.getDisDimProperty().set(PolyChart.DISDIM.TwoD);
+            } else {
+                chart.getDisDimProperty().set(PolyChart.DISDIM.OneDX);
```
JannalieT, line 575: currently this doesnt update the Traces/Contours dropdown in the spectrum status bar, but in NMR-5731, I have bound the values, so the status bar will automatically update the values so didn't worry about manually changing it here

nmrfx-bom/pom.xml
```diff
@@ -138,7 +138,7 @@
             <dependency>
                 <groupId>org.controlsfx</groupId>
                 <artifactId>controlsfx</artifactId>
-                <version>11.1.1</version>
+                <version>11.1.2</version>
                 <!-- license: BSD (3 clause) -->
```
JannalieT, line 142: This new version fixes a NPE during table filtering

JannalieT:
New changes:
- Load from datasets automatically assumes its an arrayed experiment
- Scan directory checks for unique dataset name before assuming arrayed.
- Update controlsfx version to fix table filtering NPE
- Add ability to change nFreqDims in datasets controller

vfiack (commented):
Seems to work, but date is "0" at least for RS2DData.

I'm not sure that we have the date in a header param  (it's in Serie.xml instead), but we could maybe use the file date?

Also, proc data only have the FID button enabled.

JannalieT:
> Seems to work, but date is "0" at least for RS2DData.
> 
> I'm not sure that we have the date in a header param (it's in Serie.xml instead), but we could maybe use the file date?
> 
> Also, proc data only have the FID button enabled.
Jira issues created to fix the date NMR-6013 and the FID button/dataset button on dataset browser NMR-6015

JannalieT:
Get the acquisition date from the header file, related pr: https://github.com/nanalysis/spinlab-dataset/pull/6

vfiack:
> Get the acquisition date from the header file, related pr: [nanalysis/spinlab-dataset#6](https://github.com/nanalysis/spinlab-dataset/pull/6)

I have merged that PR, removed the branch, added the binary to the public maven repository, and restarted the build here. It passes now.

brucejohnson (approved):

PR #204, opened 2022-11-07
NMR-5935: set the param STATE in the header when saving RS2D dataset

JannalieT:
Header file contents for STATE param set from NMRFx
image

Note: this pr based off NMR-5917 because of update to spinlab-dataset

### DIFF
--- a/nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/rs2d/RS2DData.java
+++ b/nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/rs2d/RS2DData.java
@@ -166,7 +166,8 @@ public class RS2DData implements NMRData {
             dataset.setSw(i, getSW(i));
             dataset.setRefValue(i, getRef(i));
             dataset.setRefPt(i, dataset.getSizeReal(i) / 2.0);
-            dataset.setFreqDomain(i, true);
+            // State can be used to set the freq domain
+            dataset.setFreqDomain(i, getState(i) == 1);
             String nucLabel = getTN(i);
             dataset.setNucleus(i, nucLabel);
             dataset.setLabel(i, nucLabel + (i + 1));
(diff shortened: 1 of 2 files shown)


nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/rs2d/RS2DData.java
```diff
@@ -1241,6 +1245,25 @@ public void setHeaderPhases(Dataset dataset) throws XPathExpressionException {
...
+            stateParam.setUuid(UUID.randomUUID().toString());
+            stateParam.setNumberEnum(NumberType.Integer);
+            stateParam.setMinValue(Integer.MIN_VALUE);
+            stateParam.setMaxValue(Integer.MAX_VALUE);
+            Order order = Stream.of(Order.values()).filter(o -> o.toString().equals(dataset.getNDim() + "D")).findFirst().orElse(Order.Unknown);
+            stateParam.setOrder(order);
+        }
+        List<Number> stateValues = new ArrayList<>(Collections.nCopies(4, 0));
+        for (int i = 0; i< dataset.getNDim(); i++) {
+            stateValues.set(i, 1);
```
vfiack, line 1261: This isn't really correct. It should contain 1 in dimensions where a fourier transform was done, and 0 in the others. I'm not sure if the dataset contains this information? Otherwise we'll keep your approximation.

vfiack: One example is pseudo 2D (T1, T2, ...), where FT is done in the first dimension only.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/rs2d/RS2DData.java
```diff
@@ -162,7 +166,8 @@ public Dataset toDataset(String datasetName) throws IOException {
             dataset.setSw(i, getSW(i));
             dataset.setRefValue(i, getRef(i));
             dataset.setRefPt(i, dataset.getSizeReal(i) / 2.0);
-            dataset.setFreqDomain(i, true);
+            // State can be used to set the freq domain
+            dataset.setFreqDomain(i, getState(i) == 1);
```
JannalieT, line 170: In Bruker, JCAMP and RS2D toDataset methods, the freq domain was just being set automatically to true. I changed here to use the state parameter to set the freq domain, but I am not sure if all processed spinit files will have this STATE parameter? If not maybe I should change back to default of true?

vfiack: They are supposed to, at least if a FT was done. If it isn't present, we can fallback to 0.

JannalieT: Okay, the getState method does return 0 if STATE isn't present and has a log.debug message for that case as well

vfiack (approved):
Code looks good. I didn't test it but I'm sure you did!

PR #208, opened 2022-11-15
Nmr 5978 move spectrum attributes controls to side pane

brucejohnson:
Not ready to merge till there is more feedback


nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/DatasetView.java
```diff
@@ -0,0 +1,214 @@
...
+                    }
+                }
+                moveToItems.add(index, startItem);
+            }
+            refreshAction();
+        }
+    }
+
+    private void refreshAction() {
+    }
```
JannalieT, line 212: why is this empty

nmrfx-processor-gui/src/main/resources/fxml/AttributesController.fxml
```diff
@@ -0,0 +1,420 @@
...
+                            </GridPane>
+                            <GridPane BorderPane.alignment="CENTER" GridPane.rowIndex="0"
+                                      GridPane.columnIndex="0" GridPane.columnSpan="1">
+                                <columnConstraints>
+                                    <ColumnConstraints minWidth="50.0" prefWidth="50.0"/>
+                                    <ColumnConstraints minWidth="50.0" prefWidth="50.0"/>
+                                    <ColumnConstraints minWidth="50.0" prefWidth="50.0"/>
+                                </columnConstraints>
+                                <children>
+                                    <Label text="X Oƒfset" GridPane.columnIndex="0" GridPane.rowIndex="0"/>
```
JannalieT, line 291: for some reason one of the f was fancy 😆 
```suggestion
                                    <Label text="X Offset" GridPane.columnIndex="0" GridPane.rowIndex="0"/>
```

nmrfx-processor-gui/src/main/resources/fxml/AttributesController.fxml
```diff
@@ -0,0 +1,420 @@
...
+                                    <Slider fx:id="xOffsetSlider" blockIncrement="0.1" max="1.0" value="0.5"
+                                            orientation="VERTICAL"
+                                            GridPane.columnIndex="0" GridPane.rowIndex="1" GridPane.columnSpan="1"
+                                    />
+                                    <Label text="Y Oƒfset" GridPane.columnIndex="1" GridPane.rowIndex="0"/>
```
JannalieT, line 296: same here
```suggestion
                                    <Label text="Y Offset" GridPane.columnIndex="1" GridPane.rowIndex="0"/>
```

JannalieT (commented):
Mostly formatting issues with a few other comments.

Tested and most things seem to be working well. Some issues are:
- in Peaks menu, if attributes is open and you pick the peaks, the peaks are not updated automatically ex. Show Peaks remains unchecked
- Apply to SCENE doesn't seem to be working for everything, exampe had 2 charts displayed in one window, switched the spectrum colour, only the active chart changed until switching the Attributes toggle
- SCENE should probably be changed to WINDOW or VIEW since scene is a javafx term people may not be familiar with
- Contour Attributes/Trace Attributes may need some thought for Apply to Scene cases for when multiple charts are displayed but have different dimensions, as it will display different options in the view
- Updating curves/contours does not update View oneD/TwoD (Note: fixed this in my NMR-5731 branch)
- Loading an FID opens the processor controller, using the Content-> Datasets tab can remove the FIID and add a dataset instead...the processor controller remains open for the content from the previous FID (probably a bug that exists in master as well)

nmrfx-processor-gui/src/main/resources/fxml/AttributesController.fxml
brucejohnson: That's my cast bumping the Alt key when I type

JannalieT (commented):
Tested, seems to be working well, just two issues:
- Attributes content overflowing when Contour Appearance or Slices accordian tab is opened
- Changing color of a projection seems a bit wonky

JannalieT (approved):

PR #210, opened 2022-11-21
Nmr 6008 spinit ppm issues

vfiack:


nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/rs2d/RS2DData.java
```diff
@@ -163,7 +159,7 @@ public Dataset toDataset(String datasetName) throws IOException {
         dataset.newHeader();
         for (int i = 0; i < nDim; i++) {
             resetSW(i);
-            dataset.setComplex(i, i == 0);
+            dataset.setComplex(i, isComplex(i));
```
vfiack, line 162: Unrelated change, but this looked like a mistake.

JannalieT (approved):
tested by changing processing type to SPINit, processing a dataset and the spectral width remained the same,

PR #211, opened 2022-11-23
Nmr 5731 drag and drop higher dim onto lower dim datasets

JannalieT:
Note: this pr will likely have significant merge conflicts with NMR-5978

### DIFF
--- a/nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/PolyChart.java
+++ b/nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/PolyChart.java
@@ -210,7 +210,7 @@ public class PolyChart extends Region implements PeakListener {
         initChart();
         drawPeaks = new DrawPeaks(this, peakCanvas);
         setVisible(false);
-
+        disDimProp.addListener((observable, oldValue, newValue) -> updateAxisType(true));
     }
 
     /**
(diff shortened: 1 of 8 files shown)


nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/PolyChart.java
```diff
@@ -2514,6 +2554,55 @@ boolean drawDatasets(GraphicsContextInterface gC) throws GraphicsIOException {
...
+        if (attributes.size() < 2) {
+            return;
+        }
+        int[] matchedDims;
+        Iterator<DatasetAttributes> attributesIterator = attributes.iterator();
+        DatasetAttributes firstAttr = attributesIterator.next();
+        while (attributesIterator.hasNext()) {
+            DatasetAttributes datasetAttributes = attributesIterator.next();
+            matchedDims = firstAttr.getMatchDim(datasetAttributes, true);
+            if (matchedDims[firstAttr.getDim(0)] == -1 || (matchedDims.length > 1 && matchedDims[firstAttr.getDim(1)] == -1)) {
```
brucejohnson, line 2574: if (matchedDims[0] == -1 || (matchedDims.length > 1 && matchedDims[1] == -1)) {

With this PR you can no longer display 2D dataset on some 3D datasets.  For example, an hsqc with dimension labels 1H and 15N will not display on an hnco with labels 1H,13C,15N (when displayed with 1H on x axis and 15N on y axis)

JannalieT: Changed to not use the getMatchDim method, since it doesn't work well when the axis were switched, now check the first attributes nucleus names and compare, was able to drag the hnco onto the hsqc, change the axis to 1H and 15N and both show up (more visible when the plane is adjusted to centre)

brucejohnson:
If you display a 2D (or higher dim) spectrum and change the display dimension (like put N on X-axis and H on y-axis for an HN hsqc spectrum) and then click the Refresh button on the spectrum attributes dialog, then the chart reverts to the default orientation.

brucejohnson (approved):

PR #213, opened 2022-11-28
test commit for testing lfs caching

JannalieT:


JannalieT:
didn't merge the github actions changes from master before making the pr, so the first run did not use the cache, the second run did use the cache but since unsure if it was just because two builds had been done in this branch, made another branch that had the github actions changes in it already: https://github.com/nanalysis/nmrfx/pull/214

PR #215, opened 2022-12-07
NMR-5951 freq domain not set correctly

JannalieT:
The freqDomain array was being set to all true for all datasets when they were loaded

### DIFF
--- a/nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/Dataset.java
+++ b/nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/Dataset.java
@@ -50,6 +50,7 @@ import java.nio.ByteOrder;
 import java.nio.file.Files;
 import java.util.*;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 /**
  * Instances of this class represent NMR datasets. The class is typically used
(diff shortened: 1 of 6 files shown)


nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/Dataset.java
```diff
@@ -111,23 +112,25 @@ public Dataset(String fullName, String name, boolean writable, boolean useCacheF
...
                 setFreqDomain(i, true);
             }
         }
 
+        DatasetParameterFile parFile = new DatasetParameterFile(this, layout);
+        parFile.readFile();
+        if (layout != null) {
+            createDataFile(raFile, writable);
+        }
+
```
JannalieT, line 133: The older NMRView files don't have freq domain set, but there wasn't a way to tell if the file was an old nv file, or if the file had just been processed without having any fourier transforms, so switched the ordering of when the parameter file was read, otherwise setting to true by default will overwrite the fdomain value in the par file.

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/bruker/BrukerData.java
```diff
@@ -203,6 +203,7 @@ public Dataset toDataset(String datasetName) throws IOException {
             dataset.setLabel(i, nucLabel + (i + 1));
             dataset.syncPars(i);
         }
+        dataset.setNFreqDims(dataset.getNDim());
         Integer ncProc = getParInt("NC_proc,1");
```
JannalieT, line 207: Bruker.toDataset automatically assumes all freqDomains are true, unsure if there is a way to determine from the files whether each dimensions is processed, right now, a pseudo 3D would have all 3 freqDomains set to true

nmrfx-processor/src/main/java/org/nmrfx/processor/processing/Processor.java
```diff
@@ -1435,9 +1435,14 @@ public void runProcesses() throws IncompleteProcessException {
...
-            dataset.setNFreqDims(nDimsProcessed);
+            int freqDimsProcessed = 0;
+            for (int i = 0; i< dataset.getNDim(); i++) {
+                if (dataset.getFreqDomain(i)) {
+                    freqDimsProcessed++;
+                }
+            }
+            dataset.setNFreqDims(freqDimsProcessed);
             for (int i = nDimsProcessed; i < dataset.getNDim(); i++) {
-                dataset.setFreqDomain(i, false);
```
JannalieT, line 1440: Not a hundred percent sure on this change or how to test, but it seems like the freqDomain is being set correctly by the FT/IFT operations, so not sure why we are setting it here, it would be good to have an example of when nDimsProcessed would be less than the dataset dimensions

brucejohnson: nDimsProcessed would be less than dataset dimensions for experiments like t1/t2 pseudo-3D, but I think its correct that that line is not necessary

brucejohnson (approved):

brucejohnson (approved):

PR #216, opened 2022-12-16
Fix infinite loop where updateAxisType called full

brucejohnson:


JannalieT (approved):
Tested and the stack overflow error no longer occurs

PR #217, opened 2022-12-21
Nmr 5977 support stack plots

brucejohnson:


JannalieT (commented):
Tested and the stacking seems to be working, there is some strange behaviour for multiple datasets in the same chart, also the vertical scale/ offset in the text fields do not adjust when the user adjusts them on the chart with their mouse.

JannalieT (approved):

PR #218, opened 2022-12-22
Nmr 6110 combine open fid dataset

JannalieT:


nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/rs2d/RS2DData.java
```diff
@@ -121,16 +121,12 @@ public class RS2DData implements NMRData {
     double tempK = 298.15;
 
     public RS2DData(String path, File nusFile) throws IOException {
-        this(path, nusFile, false);
-    }
-
-    public RS2DData(String path, File nusFile, boolean processed) throws IOException {
```
JannalieT, line 127: Relied on whether the user had wanted to open an FID or a Dataset, changed to rely on the isFID method, so if the first dimension has state = 1, then processed will be true

brucejohnson (approved):

PR #226, opened 2023-01-20
Nmr 6245 store molecule info in project

brucejohnson:


JannalieT (approved):
tested by opening 2 molecules with a project and then saving with one molecule on the chart and then creating a new project with 2 molecules and then saving the other molecule on the chart. Closing and opening the first project and adding the molecule had the correct active molecule, doing the same for the second was also was correct. Can't currently test active molecules with multiple projects open as not currently a feature.

PR #227, opened 2023-01-25
Process with Process button doesn't reload dataset

brucejohnson:


JannalieT (approved):
With RS2D and JCAMP files, clicking wildly back and forth between the process button and the autoprocesses can lead to the ocassional channel closed log message, but the spectrum still is displayed correctly so approving as discussed on teams

PR #228, opened 2023-01-27
Nmr 6285 loading some sparky list files fails

brucejohnson:


JannalieT (approved):
the two files provided in the jira issue loaded without any issues

PR #229, opened 2023-02-02
Nmr 6324 methyl atom chemical shifts dont get set when reading bmrb file

brucejohnson:


JannalieT (approved):
tested with the provided file, and the ppm is now present for the *.HB1 rows, and it matches the values in the files

PR #230, opened 2023-02-02
Nmr 6325 dont save and reload data files during processing

brucejohnson:


JannalieT (approved):
tested clicking process/autoprocess fast, clicking exit immediately after completion, loading autoprocessed vs processed files, opening a file that would autoprocess in another window while one is still processing, processed files showing up in RecentFiles, tested opening from SPINit in windows

PR #231, opened 2023-02-02
Nmr 6320 update attributes display when different dataset item is selected

brucejohnson:


JannalieT (approved):
tested with 2 datasets, was able to set each dataset to have postives and negatives with a different color, switching between datasets changed the attributes panel to the correct negative/positive contour checkbox options and color for each dataset

PR #233, opened 2023-02-03
NMR-4777 mutalize mousebind action between spinit and NMRF

JannalieT:
Change to scale with scroll wheel instead of ctrl+scroll, assume mac use natural scrolling and windows and linux use traditional scrolling, for 1D left border will change vertical offset instead of zoom

Slightly different from Jira issue 
-> double click will not zoom out (@brucejohnson requested this not be implemented)
-> Scale will adjust on scroll instead of ctrl + scroll (does the same thing whether above or below the baseline)

### DIFF
--- a/nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/spectra/GestureBindings.java
+++ b/nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/spectra/GestureBindings.java
@@ -21,6 +21,7 @@ import javafx.event.Event;
 import javafx.scene.input.RotateEvent;
 import javafx.scene.input.ScrollEvent;
 import javafx.scene.input.ZoomEvent;
+import org.apache.commons.lang3.SystemUtils;
 import org.nmrfx.processor.gui.PolyChart;
 
 /**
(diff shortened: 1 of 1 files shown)


nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/spectra/GestureBindings.java
```diff
@@ -74,15 +75,17 @@ public void scroll(ScrollEvent event) {
         ChartBorder border = chart.hitBorder(x, y);
         double dx = event.getDeltaX();
         double dy = event.getDeltaY();
-        if (event.isControlDown() && border == ChartBorder.NONE) {
-            chart.scaleY(dy);
-        } else if (border == ChartBorder.RIGHT || border == ChartBorder.TOP) {
-            chart.updateProjectionScale(border, dy);
+        int scrollDirectionFactor = SystemUtils.IS_OS_MAC  ? 1 : -1;
```
JannalieT, line 78: This assumes all Macs will use natural scrolling, and Windows and Linux will use traditional scrolling... both Mac and Linux can change the scroll direction in their settings...I couldn't find any way to detect this from java, if we want to add a setting for this, we could have one to switch our assumption of the scrollDirectionFactor within NMRFx but we couldn't specify if its natural of traditional scrolling direction...so didn't add a setting for this at this time

brucejohnson (approved):

PR #237, opened 2023-02-10
Nmr 4492 unable to process data in folder containing special characters

JannalieT:
In many places we pass filenames as strings from Java to Python to Java, and the special characters can get messed up as the encoding doesn't seem to be the same. Tried to set the encoding in VM arguments using -Dpython.console.encoding=UTF-8 and -Dfile.encoding=UTF-8 which did not work. So tried to make sure filepaths were either passed to the PythonInterpreter as a variable using `set` or passing strings containing file paths to python as unicode string and passing the strings back to java by decoding them. Also fixes the special characters not displaying correctly in the NMRFx Console, by storing character bytes in an array and writing them all at once when flush is called instead of 1 by 1.


brucejohnson:
Invalid commands in console used to show an error message, but now they don't and you have to hit the Clear button to recover.
Previously, typing "who", for example would give this:
> who
Traceback (most recent call last):
  File "<input>", line 1, in <module>
NameError: name 'who' is not defined
> 
and you could type more commands,
but now you don't see anything in the NMRFx console, in the terminal you'll see something like:
Exception in thread "JavaFX Application Thread" Traceback (most recent call last):
  File "<string>", line 1, in <module>
NameError: name 'who' is not defined

brucejohnson (approved):

PR #245, opened 2023-02-23
Nmr 5648 autophase can invert spectrum

brucejohnson:


JannalieT (approved):
tested with FID from jira issue as well as a few other datasets including one with a strong dataset, changing the parameters for autophase (winSize and ratio) the spectrum was always displayed right side up

PR #246, opened 2023-02-23
Allow autophase from phase menu in spectrum mode

brucejohnson:


nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/PolyChart.java
```diff
@@ -1455,19 +1455,33 @@ protected void setPhasePivot() {
     }
 
     protected void autoPhaseFlat(boolean doFirst) {
-        DatasetBase dataset = getDataset();
```
JannalieT, line 1458: This function should be renamed to autoPhaseLeastNeg I think

JannalieT (commented):
tested and works well, just the one comment needs to be fixed, sonar lint is complaining about duplicated code, the first half of autoPhaseMax and autoPhaseFlat are the same and could be combined into a function that returns VecBase

JannalieT (approved):

PR #251, opened 2023-02-28
NMR-6439: when updating dataset, if name matches previous dataset, us…

JannalieT:
…e the same axes dims


nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/PolyChart.java
```diff
@@ -1875,12 +1875,14 @@ public DatasetAttributes setDataset(DatasetBase dataset, boolean append, boolean
...
                     DatasetBase existingDataset = datasetAttributes.getDataset();
                     double oldLevel = datasetAttributes.getLvl();
+                    int[] oldDims = datasetAttributes.getDims();
                     datasetAttributes.setDataset(dataset);
                     if ((existingDataset == null) || (!keepLevel && !existingDataset.getName().equals(dataset.getName()))) {
                         datasetAttributes.setLvl(dataset.getLvl());
                     } else if ((existingDataset != null) && existingDataset.getName().equals(dataset.getName())) {
                         datasetAttributes.setLvl(oldLevel);
                         datasetAttributes.setHasLevel(true);
+                        datasetAttributes.setDims(oldDims);
```
JannalieT, line 1885: Assuming here that if the existing dataset name matches dataset name they will have the same number and nuclei dimensions. Is this okay, or need to add a further check?

brucejohnson (approved):

PR #253, opened 2023-03-02
Nmr 6458 varian array data isnt currently supported

brucejohnson:


JannalieT (approved):
tested with provided file and can change display mode to stack plot and there were 11 spectrums stacked

PR #254, opened 2023-03-02
Nmr 6415 varian files with f 1 coefs 1 0 0 1 not processed correctly

brucejohnson:


nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/varian/VarianData.java
```diff
@@ -696,20 +708,24 @@ public void setArraySize(int iDim, int size) {
...
             return !(s != null && s.equals("rft")); // proc="ft" or "lp"
         } else {
-            String ext = "";
-            if (iDim > 1) {
-                ext += iDim;
-            }
-            String s = getPar("phase" + ext);
+            String ext = String.valueOf(iDim);
+            String s = getPar("proc"+ext);
+            boolean notRFT =  !(s != null && s.equals("rft"));
```
JannalieT, line 716: This is a bit confusing to read, I think this would be equivalent and will still check the null value
```suggestion
            boolean notRFT =  !"rft".equals(s); 
```

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/varian/VarianData.java
```diff
@@ -666,6 +665,19 @@ public final int getSize(int iDim) {
         return sizes[iDim];
     }
 
+    private int getNI(int i) {
```
JannalieT, line 668: maybe a doc string here? what is NI stand for

JannalieT (commented):
Tested with provided file, loading the file the generated script contained a FT for both dimensions and the first dimension had a COMB item added on at the end of the first dimension operations

PR #257, opened 2023-03-06
NMR-6466: create method to copy table data to tab seperated string an…

JannalieT:
…d add copy table button and command + c keybinding to PeakTable

Tested with Excel on Windows and LibreOffice Calc on ubuntu, need testing on spreadsheet in macos

### DIFF
--- a/nmrfx-utils/src/main/java/org/nmrfx/utils/TableUtils.java
+++ b/nmrfx-utils/src/main/java/org/nmrfx/utils/TableUtils.java
@@ -1,0 +1,48 @@
+package org.nmrfx.utils;
+
+import javafx.scene.control.TableColumn;
+import javafx.scene.control.TableView;
+import javafx.scene.input.Clipboard;
+import javafx.scene.input.ClipboardContent;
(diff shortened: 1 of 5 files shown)


nmrfx-utils/src/main/java/org/nmrfx/utils/TableUtils.java
```diff
@@ -0,0 +1,48 @@
...
+    private TableUtils() {}
+
+    /**
+     * Copies the contents of a table, including the headers, to the clipboard as a tab separated string.
+     * If no rows are selected or copyAll is true, then all rows will be copied. Otherwise only the
+     * selected rows will be copied.
+     * @param tableView The TableView to copy.
+     * @param forceCopyAll If true, all rows will be copied regardless of selections
+     */
+    public static void copyTableToClipboard(TableView<?> tableView, boolean forceCopyAll) {
```
JannalieT, line 24: Used wildcard here so we could use this function for tables other than the Peak table.
Have tested with the peak table, region table, log table and change log level table

brucejohnson (approved):

PR #259, opened 2023-03-07
Nmr 6477 add filter on peak status to peak tool

brucejohnson:


JannalieT (approved):
Tested with runabout5 that had multiple peak lists, the selecting a filter, only the peaks in that list are displayed (if peak list table says 3 partial peaks, only 3 peaks can be scrolled though in the Peak Tool)
When a filter is selected and the peak list is changed, the filter is reset and the displayed peak is reset
When a filter is selected and a peak is opened from the peak table, the filter is reset and the selected peak is displayed

PR #261, opened 2023-03-08
NMR-6498-py-structure-test-fails-sometimes

brucejohnson:
Raise energy threshold for passing
Random structures could fail with previous structures


JannalieT (approved):
tested on linux vm running the two ways it was consistently failing previously and the tests passed

PR #264, opened 2023-03-17
Nmr 6579 indexing for non phase sensitive files

brucejohnson:


nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/varian/VarianData.java
```diff
@@ -762,6 +766,30 @@ public boolean isComplex(int iDim) {
         }
     }
 
+    @Override
+    public int getGroupSize(int iDim) {
+        if (iDim == 0) {
+            String s = getPar("proc");
+            return !(s != null && s.equals("rft")) ? 2 : 1; // proc="ft" or "lp"
```
JannalieT, line 773: I think this is a bit complicated to read with all the negatives, just double check that I have adjusted it correctly so that if s is rft then the group size is 1, otherwise it is 2. 
```suggestion
            return "rft".equals(s) ? 1 : 2; // proc="ft" or "lp"
```

nmrfx-processor/src/main/java/org/nmrfx/processor/datasets/vendor/NMRData.java
```diff
@@ -325,16 +319,20 @@ default public void setFixDSP(boolean value) {
...
      * @param dim data dimension index
      * @param value if data is complex
      */
-    public default void setComplex(int dim, boolean value) {
+    default void setComplex(int dim, boolean value) {
+
+    }
 
+    default int getGroupSize(int dim) {
+        return isComplex(dim) ? 2 : 1;
```
JannalieT, line 335: add doc string here

PR #283, opened 2023-05-09
Dataset doc

vfiack:
A few notes to document the various classes related to loading datasets and their relation.

The plantuml diagram looks like this:
image

### DIFF
--- a/docs/dataset.md
+++ b/docs/dataset.md
@@ -1,0 +1,134 @@
+# Dataset related classes
+
+## NV file format
+
+The NV file format is described in [nv-file-format](nv-file-format.md).
+
+This documentation is slightly outdated, a few fields have been added to the header.
+It remains a good introduction to understand the file layout.
+
+## Class diagram
+
+```plantuml
+@startuml
+
+package org.python.core {
+    class PySequence {}
+}
+
+package core {
+    class DatasetBase {
+        DatasetLayout layout;
+        VecBase vecMat;
+        List<DatasetRegion> regions;
+        DatasetStorageInterface dataFile;
+    }
+
+    class DatasetLayout {
+       block sizes and offsets
+    }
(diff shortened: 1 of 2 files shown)


docs/dataset.md
```diff
@@ -0,0 +1,134 @@
...
+### Subclassing could be simplified
+
+* Dataset is the only subclass of DatasetBase. There is no clear reason to keep them separated if we have them in the
+  same maven module.
+  * DatasetBase instances are created by:
+    * NMRStarReader. However, the instance is discarded directly after creation! Unless there is a hidden side-effect,
+      this has no use and could be removed.
+    * DatasetFactory only if Dataset is not found (should never happen when running processor or analyst)
+* Vec is the only subclass of VecBase. Same issue as between Dataset and DatasetBase
+  * VecBase constructors are only called from Vec. There is no direct usage.
```
vfiack, line 103: @brucejohnson can you confirm that those are the only usage? Or is there a weird case I missed that would create a DatasetBase or VecBase instance from python for example?

docs/dataset.md
```diff
@@ -0,0 +1,134 @@
...
+  BigMappedMatrixFile.
+  * this ignores the file header and block padding, this isn't the full file size.
+* otherwise, if the layout correspond to a sub-matrix (true every time more than one block is defined), use
+  MappedSubMatrixFile.
+* finally, fallback to MappedMatrixFile.
+
+I am not sure if MappedMatrixFile is really necessary. Is there's really a performance gain compared to
+using MappedSubMatrix file even where there is a single block?
+
+I am also not sure why using SubMatrixFile when processing on windows instead of MappedSubMatrixFile?
```
vfiack, line 128: @brucejohnson do you have more information on these two questions?

PR #288, opened 2023-05-10
NMR-6877: remove commented out code - nmrfx-core

vfiack:


nmrfx-core/src/main/java/org/nmrfx/datasets/DatasetHeaderIO.java
```diff
@@ -317,7 +317,7 @@ public final synchronized DatasetLayout readHeaderUCSF(RandomAccessFile raFile)
             lay.dimDataset();
             d.setDataType(0);
         } catch (IOException e) {
-            //LOGGER.error("Can't read header ", e);
+            log.error("Can't read header ", e);
```
vfiack, line 320: slight behavior change here: since we now have a logging system, I've opted to put the log again instead of removing the line.

PR #295, opened 2023-05-12
Nmr 6885 remove unused classes

vfiack:


docs/removed-code.md
```diff
@@ -0,0 +1,23 @@
+# Removed code
```
vfiack, line 1: @brucejohnson is this the kind of file you were hoping for?

PR #296, opened 2023-05-12
Nmr 6885 remove unused methods - processor-gui - partial

vfiack:
Only a few easy ones to start with.


nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/ChartProcessor.java
```diff
@@ -936,77 +887,6 @@ public String buildScript(int nDim) {
         return scriptBuilder.toString();
     }
 
-    public String buildMultiScript(String baseDir, String outputDir, ArrayList<String> fileNames, boolean combineFiles) {
```
vfiack, line 939: @brucejohnson I'm not sure about his one

nmrfx-processor-gui/src/main/java/org/nmrfx/processor/gui/controls/ConsoleUtil.java
```diff
@@ -74,123 +35,4 @@ public static void runOnFxThread(final Runnable runnable) {
             Platform.runLater(runnable);
         }
     }
-
-    public void banner() {
```
vfiack, line 78: Surprisingly enough, nothing from this class is actually used in the console.

PR #299, opened 2023-05-12
NMR-6891: regroup processor & analyst, keep only analyst

vfiack:
No code change, only moving files & updating poms.


nmrfx-analyst/pom.xml
```diff
@@ -13,11 +13,6 @@
     <artifactId>nmrfx-analyst</artifactId>
     <packaging>jar</packaging>
 
-    <properties>
-        <mainClass>com.onemoonsci.datachord.nvfx.MainApp</mainClass>
-    </properties>
```
vfiack, line 18: This was wrong and unused

PR #305, opened 2023-05-16
Nmr 6893 split fxml controller

vfiack:
Trying to progressively reduce FXMLController's responsability.

Moving instance management to a separate FXMLControllerManager.
Classes should be renamed, but at a later time since some plugins reference them.

Most file changes are very simple and only update the calling path.
Main changes that need review are in FXMLController & FXMLControllerManager.

### DIFF
--- a/nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/FXMLControllerManager.java
+++ b/nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/FXMLControllerManager.java
@@ -9,7 +9,6 @@ import javafx.stage.StageStyle;
 import org.nmrfx.analyst.gui.AnalystApp;
 
 import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
 import java.io.IOException;
 import java.util.Collection;
 import java.util.Collections;
@@ -20,9 +19,11 @@ import java.util.Set;
  * Manages instances of FXMLControllers, from creation to listing them and keeping a way to know which one is the active one.
  */
 public class FXMLControllerManager {
+    // The active controller is the last one created, or the one that has focus
     private final SimpleObjectProperty<FXMLController> activeController = new SimpleObjectProperty<>(null);
 
-    // it is important to keep insertion order: the first controller is considered the main one and is kept when closing all secondary stages
+    // A collection of all known controllers.
+    // It is important to keep insertion order: the first controller is considered the main one and is kept when closing all secondary stages
     private final Set<FXMLController> controllers = new LinkedHashSet<>();
 
     public Collection<FXMLController> getControllers() {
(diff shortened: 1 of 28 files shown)


vfiack:
I'll ignore sonar warning here for now. It was already an issue before, I only made it more obvious.

nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/FXMLControllerManager.java
```diff
@@ -0,0 +1,163 @@
...
+    /**
+     * Set the active controller or null.
+     *
+     * @param active the active controller.
+     * @throws IllegalStateException if the controller is not registered
+     */
+    public void setActiveController(@Nullable FXMLController active) {
+        if (active != null && !controllers.contains(active)) {
+            throw new IllegalStateException("Trying to set an unregistered controller as the active one!");
+        }
```
JannalieT, line 105: Encountered this exception following these steps
1. Start NMRFx
2. Go to Spectra -> New Window
3. Close original window
4. Go to Projects -> Save and save project
5. Check logs for exception

nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/FXMLControllerManager.java
```diff
@@ -0,0 +1,163 @@
...
+     * Get the active controller, creating one if necessary.
+     *
+     * @return the active controller.
+     */
+    @Nonnull
+    public FXMLController getOrCreateActiveController() {
+        FXMLController active = activeController.get();
+        if (active == null) {
+            //XXX is there actually a case where not having an active controller is a possibility?
+            active = newController();
```
JannalieT, line 77: I'm not sure if there is, but one thing to check would be to run on mac os. Since I think the File bar can occur separately, I'm not sure if its possible to have all the windows closed but still have the program running with the file menu bar.

JannalieT (commented):
Tried creating multiple windows and closing them in different orders, encountered two exceptions related to saving projects. Exception 1 (uncaught) steps to reproduce in comments

Exception 2 - caught exception that I haven't seen before:
1) Start NMRFx
2) Go to Projects -> Save
3) See stack trace
[NoSuchMethodException.txt](https://github.com/nanalysis/nmrfx/files/11492561/NoSuchMethodException.txt)

vfiack:
> Tried creating multiple windows and closing them in different orders, encountered two exceptions related to saving projects. 

Thanks a lot for these tests! I've missed both errors.

I was kind of expecting #1 when I decided to throw this exception; I'll likely replace it by a log but it's indicative of bug somewhere else. 

I wasn't expecting #2. It looks like I remove a field that was used from a python script. I'll double check to see if I can find others.

nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/FXMLControllerManager.java
vfiack: Found and fixed the root cause: GUIScripter keeps a FXMLController instance in a static field, but isn't notified when the window is closed and unregistered from the instance list.

This causes a temporary memory leak (the controller should be garbaged but will not be until another one is set to GUIScripter) and the bug you found. I've kept the static field but replaced by the active one when calling `GUIScripter.getControlller()` if it has been unregistered.

I've kept the exception in place in FXMLController manager since I believe it can help us find other bugs similar to this one.

vfiack:
> Exception 2 - caught exception that I haven't seen before

Found the root cause for this one as well. I had kept the fields in FXMLController, but didn't keep the public get/set accessors, since they looked unused. However, they are actually used by reflection internally from PropertyUtils methods. This will be a trap for other cleanup efforts.

I restored the setters/getters here, but we will need to try to get rid of PropertyUtils and explicitly expose which fields should be publicly exposed.

JannalieT (approved):
tested again and exceptionless

PR #306, opened 2023-05-16
Nmr 6893 simpler fxml loader

vfiack:
Simplify loading FXML files, remove all copy/pasted code (sometimes with wrong class name to access resources).
Ensure stage is always set when creating controller that need it.


JannalieT (approved):
Opened various gui components and they loaded okay

PR #308, opened 2023-05-17
Nmr 6904 explicit config fields

vfiack:
Refactor accessing properties from string, expose them explicitly, remove dependency to beanutils.

This should prevent errors like the one found in https://github.com/nanalysis/nmrfx/pull/305#pullrequestreview-1429523377 in the future.

We lose lazy-init for these properties, but
1. I think it's worth it, and
2. Properties where being accessed in any case, so they would likely most be created anyway.


JannalieT (approved):
Tested by changing some of the dataset attributes for the chart, everything seems to be working.

Did notice one thing that if you go to Attributes -> Colors -> and check "Cross 0" it changes the color of Cross 0 to the specified color...and Cross 1 changes to magenta for some reason, this happens in master too and not sure if it is supposed to be a feature or not, since unchecking Cross 0 will change it back to red...

vfiack:
Probably a bug, I don't see why magenta would be a special case that anyone would want. I'll still go ahead and merge since it was already present before.

PR #312, opened 2023-05-26
Nmr 6922 polychart instance management

vfiack:
Not finished, but if we can already test and merge that, it's better than nothing

### DIFF
--- a/nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/GUIScripter.java
+++ b/nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/GUIScripter.java
@@ -23,7 +23,6 @@ import java.util.concurrent.FutureTask;
 import java.util.stream.Collectors;
 
 /**
- *
  * @author Bruce Johnson
  */
 @PythonAPI("gscript")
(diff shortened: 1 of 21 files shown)


nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/GUIScripter.java
```diff
@@ -71,7 +71,8 @@ PolyChart getChart() {
     }
 
     public String active() {
-        PolyChart chart = useChart != null ? useChart : PolyChart.getActiveChart();
+        PolyChart chart;
+        chart = useChart != null ? useChart : PolyChartManager.getInstance().getActiveChart();
```
JannalieT, line 75: whats the purpose of this change? just formatting to not make the line as long?

JannalieT (approved):
tested by creating multiple charts in windows and deleting spectrums, and saving and reloading projects, seems to be working well

nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/GUIScripter.java
vfiack: This wasn't done on purpose. 

This is likely a side effect of using intellij's "inline method" refactoring tool. I'll change it back manually before merging.

PR #313, opened 2023-05-30
Nmr 6922 polychart synchronization

vfiack:
Extracted synchronization features out of PolyChart, then rewrote the extracted part.

Introduced a Bound enum in Axis to avoid having meaningless 0 and 1 values in AxisChangeListener & PolyChartSynchronizer.

### DIFF
--- a/nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/spectra/SpectrumMenuActions.java
+++ b/nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/spectra/SpectrumMenuActions.java
@@ -7,6 +7,7 @@ import javafx.stage.Stage;
 import javafx.stage.StageStyle;
 import org.nmrfx.analyst.gui.AnalystApp;
 import org.nmrfx.analyst.gui.MenuActions;
+import org.nmrfx.processor.gui.PolyChart;
 import org.nmrfx.processor.gui.PolyChartManager;
 import org.nmrfx.processor.gui.controls.GridPaneCanvas;
 import org.nmrfx.processor.gui.spectra.WindowIO;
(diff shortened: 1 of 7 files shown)


JannalieT (commented):
Code looks okay, couldn't test very well because I'm not exactly sure what synchronization is supposed to be doing...
According to the NMRFx website docs:
>NMRFx provides a basic implementation of spectrum synchronization. More complete tools are in development. Synchronization is currently used to synchronize the spectrum charts contained within a single window. Just choose Sync Axes from the Spectrum Menu. Now whenever you change the X or Y axis view, or planes in one spectrum chart, the same view will be set in the other charts. Synchronization only happens for dataset axes that share the same label, but the dataset axes do not need to be displayed on the same chart axis (X, Y, Z...).

I would assume from that description that changing the axis orientation on one chart (see screenshot) would not change other charts in a multi chart view unless we have selected the `Sync Axes` option....but even without clicking `Sync Axes` the chart axes still seem to be synchronizing but the groups list is empty. So someplace else the axes are already synchronizing? or I'm missing the point of synchronization?

image

vfiack:
I didn't read the documentation, so I don't know what it was supposed to do. However, I read the previous code, so I know what it was actually doing:

* open a window,
* use a grid display, 
* open multiple datasets that share at least one nucleus. I tested with several basic proton 1D 
* use the "Sync Axes" item in "spectra" menu. This will create the sync group containing all charts displayed in the active window
* move the axis to the left/right or zoom in. All charts will see their axis change the same way.


There are several strange things however:
* if you move one chart first, then enable synchronization: axes remain different until you move one. They don't get synchronized when enabling the feature
* groups are attached to chart and not to the window, so if you later change the grid to add more datasets, only the one that were previously synched will move. New charts will remain independent in the same window.

vfiack:
> Example of synchronizing without specifying Sync Axes option:

This is done by SpectrumStatusBar.dimAction(..). I believe that was already the case before this change.

JannalieT:
> * move the axis to the left/right or zoom in. All charts will see their axis change the same way.
Ah that was what I missed doing when I was testing, I thought the axis nucleus switching was the synced part but it is not.

JannalieT (approved):

PR #316, opened 2023-05-31
Nmr 6922 polychart borders

vfiack:
Replace fields & array with Insets class. Easier to follow when each side has a name instead of an index...

Correspondence with old array indexes:
 0 = left
 1 = right
 2 = bottom
 3 = top


JannalieT (approved):
tested by changing the border dimensions for a chart.

PR #317, opened 2023-06-01
Nmr 6922 polychart axes

vfiack:
Tried to extract axis management outside of polychart, and to avoid accessing by index for x/y axis.

Not sure if this improves the situation or made it worse...
Kept the inner array exposed for now to limit the amount of changes to review at once.


JannalieT (approved):
tested by opening multiple spectrum with different number of axes (1, 2, 3)  changing the axes, dragging the axes, with and without grid mode, overlay spectrum with valid and invalid axes

PR #319, opened 2023-06-01
Nmr 6922 merge nmraxis axis

vfiack:
NMRAxis didn't add anything that was not already handled by the parent class. Removed it and use the parent class directly.

Kept NMRAxisBase and NMRAxisIO: they not directly related to this class hierarchy (which was strange: in all others cases, XXXBase is a parent of XXX, for ex: Vec, Dataset, Project, ...)


nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/PolyChart.java
```diff
@@ -1672,27 +1672,27 @@ public void updateAxisType(boolean alwaysUpdate) {
             axes.setMode(0, AXMODE.TIME);
         }
         boolean reversedAxis = (axes.getMode(0) == AXMODE.PPM);
-        boolean autoScale = reversedAxis != axes.getX().getReverse();
+        boolean autoScale = reversedAxis != axes.getX().isReversed();
         axes.getX().setReverse(reversedAxis);
         String xLabel = axes.getMode(0).getLabel(datasetAttributes, 0);
-        axes.getX().updateStateAndLabel(xLabel);
+        axes.getX().updateLabel(xLabel);
```
JannalieT, line 1678: so don't need to update the state here anymore?

JannalieT (approved):
tested by loading spectrum and switching/adjusting axes

nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/PolyChart.java
vfiack: I don't think so. 

Previously, we had two classes: NMRAxis > Axis.

NMRAxis had a flag to set ticks and label visible, but setting the flag had no effect. Then when updating the label, it also used to set the tick & label visibility depending on this flag.

When I regrouped those classes, I chose to remove this flag, and instead directly set the state in the Axis flag. There may be a side effect if someone used to set the flag in NMRAxis, then changed options in Axis and finally called NMRAxis.updateStateAndLabel(). That would be very strange, so I'm willing to take that risk.

JannalieT: Okay, sounds good, the axes seem to be working okay anyways

PR #323, opened 2023-06-01
Nmr 6922 polychart accessors

vfiack:
No direct field access anymore for PolyChart.

FXMLController was exposed by 2 different getters. I've kept only the most explicit one.

### DIFF
--- a/nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/PolyChart.java
+++ b/nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/PolyChart.java
@@ -130,11 +130,11 @@ public class PolyChart extends Region implements PeakListener {
 
     //XXX use accessor instead
     FXMLController controller;
-    ProcessorController processorController = null;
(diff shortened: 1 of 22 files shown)


nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/PolyChart.java
```diff
@@ -166,9 +161,9 @@ protected PolyChart(FXMLController controller, String name, Pane plotContent, Ca
         this.canvas = canvas;
         this.peakCanvas = peakCanvas;
         this.annoCanvas = annoCanvas;
-        plotBackground = new Group();
         this.plotContent = plotContent;
         drawSpectrum = new DrawSpectrum(axes, canvas);
+        drawSpectrum.setCancelButton(controller.getCancelButton());
```
JannalieT, line 166: Got NPE in drawSpectrum.setCancelButton because at this point the controller has not yet created the cancel button.
See lines 842 (polyChart creation) and 844 (cancelbutton creation) in FXMLController

PR #325, opened 2023-06-05
Nmr 6953 explicit shared canvas

vfiack:
A few direct access to inside canvas remains, but the PR starts to get too big. May be worth an intermediate review.

Goal of this change is to be very clear on what canvas is used to paint what components, show that they are shared, and manage them together as one object. This also removes some cross-references to fxmlcontroller hopefully.


JannalieT (approved):
tested with spectrum with peaks and molecule on it, can still select and move around peaks/molecule

PR #327, opened 2023-06-06
Nmr 6976 prototype accordion

JannalieT:
Only the ModifiableAccordion class is the prototype, the other changes are just for demo purposes to test it out.


nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/GUI/ModifiableAccordion.java
```diff
@@ -0,0 +1,97 @@
...
+        titledPane.setGraphicTextGap(0);
+        titledPane.setContentDisplay(ContentDisplay.GRAPHIC_ONLY);
+        accordion.getPanes().add(titledPane);
+
+        // Listener to set the minWidthProperty once the scene is displayed (FRAGILE!)
+        titleBox.widthProperty().addListener((obs, oldWidth, newWidth) ->{
+            if (!titleBox.minWidthProperty().isBound()) {
+                // Need to calculate how much space the arrow icon takes up once the stage is showing, can only calculate
+                // once the stage has been shown and the sizing has been calculated
+                Region arrowRegion = (Region) titledPane.lookup(".arrow-button");
```
JannalieT, line 58: this would break if someone decided to rename the arrow-button (or using our own custom css class names somehow)

nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/GUI/ModifiableAccordionScrollPane.java
```diff
@@ -0,0 +1,118 @@
...
+
+        }
+
+        @Override
+        protected void layoutChildren(final double x, final double y, final double w, final double h) {
+            super.layoutChildren(x, y, w, h);
+            double arrowWidth = arrow.getLayoutBounds().getWidth();
+            double arrowPadding = arrow.getPadding().getLeft() + arrow.getPadding().getRight();
+
+            ((Region) getSkinnable().getGraphic()).setMinWidth(w - (arrowWidth + arrowPadding));
```
JannalieT, line 115: moved what was in the listener/property binding here instead. This is much neater and seems to work well. It was still not my idea solution since it only works with ContentDisplay.GRAPHIC_ONLY, meaning I need to add a label for the title instead of using the built in text. I did attempt to calculate the length of the gap between the text and the graphic to be able to call setGraphicTextGap but I could not get the calculations to work at all, so decided to just use the GRAPHIC_ONLY display

PR #336, opened 2023-06-08
NMR-7010: review non-final static fields: core

vfiack:
Lots of code issues in `org.nmrfx.chemistry` but this was out of scope for this PR.


nmrfx-core/src/main/java/org/nmrfx/chemistry/utilities/Format.java
```diff
@@ -1,87 +0,0 @@
...
-package org.nmrfx.chemistry.utilities;
-
-import java.text.DecimalFormat;
-import java.text.NumberFormat;
-import java.util.Locale;
-
-/**
- * @author brucejohnson
- */
-public class Format {
```
vfiack, line 28: This was 100% duplicated with org.nmrfx.utilities.Format. I've kept only the other one.

nmrfx-core/src/main/java/org/nmrfx/chemistry/constraints/DistanceConstraintSet.java
```diff
@@ -31,6 +31,49 @@
...
+            "_Torsion_angle_constraint.Atom_ID_2",
+            "_Torsion_angle_constraint.Atom_type_2",
+            "_Torsion_angle_constraint.Resonance_ID_2",
+            "_Torsion_angle_constraint.Assembly_atom_ID_3",
+            "_Torsion_angle_constraint.Entity_assembly_ID_3",
+            "_Torsion_angle_constraint.Entity_ID_3",
+            "_Torsion_angle_constraint.Comp_index_ID_3",
+            "_Torsion_angle_constraint.Seq_ID_3",
+            "_Torsion_angle_constraint.Comp_ID_3",
+            "_Torsion_angle_constraint.Atom_ID_3      _Torsion_angle_constraint.Atom_type_3",
```
JannalieT, line 61: This was the previous version but it must be a typo...since looking at line 43,44,45 and line 52,53,54, @brucejohnson could confirm?

JannalieT (dismissed):
seems to be working, approving since the typo was there previously, I don't know whether we should fix here or not

nmrfx-core/src/main/java/org/nmrfx/chemistry/constraints/DistanceConstraintSet.java
vfiack: Yes, this clearly looks like a typo. I'll fix it here.

JannalieT (approved):

PR #337, opened 2023-06-08
NMR-6904: intensityAxis property name didn't match the field name.

vfiack:
This lead to an inconsistency when loading existing project files created before the change on public properties management.


JannalieT (commented):
Using runabout5 project,in my accordion ptototype branch I get a warning about the property name not being exposed but the project still loads, using this branch I don't have that warning but end up with an out of bounds exception

PR #339, opened 2023-06-08
NMR-6522: don't show console stage on creation, make sure minimized c…

JannalieT:
…onsole is shown on View -> show console

### DIFF
--- a/nmrfx-utils/src/main/java/org/nmrfx/console/ConsoleController.java
+++ b/nmrfx-utils/src/main/java/org/nmrfx/console/ConsoleController.java
@@ -22,7 +22,6 @@ import javafx.event.Event;
 import javafx.event.EventHandler;
 import javafx.fxml.FXML;
 import javafx.fxml.Initializable;
-import javafx.geometry.Rectangle2D;
 import javafx.scene.control.TextArea;
 import javafx.scene.input.KeyCode;
 import javafx.scene.input.KeyEvent;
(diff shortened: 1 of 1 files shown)


nmrfx-utils/src/main/java/org/nmrfx/console/ConsoleController.java
```diff
@@ -107,14 +107,7 @@ public static ConsoleController create(InteractiveInterpreter interpreter, Strin
         consoleController = Fxml.load(ConsoleController.class, "ConsoleScene.fxml")
                 .withStage(stage)
                 .getController();
-
-        stage.show();
-        Screen screen = Screen.getPrimary();
-        Rectangle2D screenSize = screen.getBounds();
-        stage.toFront();
-        stage.setY(screenSize.getHeight() - stage.getHeight());
```
JannalieT, line 115: I removed the code that calculates where to put the box initially because I would need to first show the stage and then hide it which will take time and also flash the console on the screen while starting up. This means that when the user clicks "Show Console" the first time it will pop up in the centre of the screen which I think should be acceptable because the user is actively looking for it, once they have moved the console where they want it, "Show Console" will pop it up where they put it. I can add this calculation back in if not acceptable

vfiack (commented):
The console is well hidden, but the software still closes when close it. 

We should remove that as well, users should be able to close/hide the console after showing it.

vfiack (approved):

PR #340, opened 2023-06-08
NMR-6968: move phasing/content/attributes toggle buttons up one menu bar

JannalieT:
For windows, the toggle buttons can be added on the same level of the file menus...but if they are put in a toolbar the buttons will have padding around them, which affects the alignment with the menu items, (increasing the menu bar height looks weird since the menu options do not increase in size) so I put the buttons in a stackpane without padding so they will align correctly to the file Menus, below is a screen shot of how that looks. If we need to have padding space around the buttons, it will become a lot more complicated....
image

The mac global menu is set using MenuToolkit which only accepts a MenuBar, buttons cannot be added to a MenuBar so had to add an extra menu for the toggle buttons for mac os. Since the buttons are the only item in the toolbar, they can be displayed with or without top/bottom padding. I did not include padding to match with the style of the windows buttons.

Note: should be tested on mac before merging


nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/FXMLController.java
```diff
@@ -863,6 +861,39 @@ public void initialize(URL url, ResourceBundle rb) {
         borderPane.heightProperty().addListener(e -> contentController.updateScrollSize(borderPane));
     }
 
+    /**
+     * Initialize the toggle buttons Phasing, Attributes and Contents. On mac these buttons will appear right
+     * aligned in a separate top menu in the window, otherwise they will appear right aligned in the file menu.
+     */
+    public void initializeToggleButtons() {
```
vfiack, line 868: method should be private and maybe have a more explicit name? There could be other toggle buttons somewhere in the UI.

nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/FXMLController.java
```diff
@@ -863,6 +861,39 @@ public void initialize(URL url, ResourceBundle rb) {
...
+        ToggleButton attributesButton = new ToggleButton("Attributes");
+        ToggleButton contentButton = new ToggleButton("Content");
+        attributesButton.setOnAction(e -> toggleSideBarAttributes(phaserButton, attributesButton, contentButton));
+        contentButton.setOnAction(e -> toggleSideBarAttributes(phaserButton, attributesButton, contentButton));
+        phaserButton.setOnAction(e -> toggleSideBarAttributes(phaserButton, attributesButton, contentButton));
+        phaserButton.getStyleClass().add("toolButton");
+        attributesButton.getStyleClass().add("toolButton");
+        contentButton.getStyleClass().add("toolButton");
+        SegmentedButton groupButton = new SegmentedButton(phaserButton, contentButton, attributesButton);
+        if (!AnalystApp.isMac()) {
```
vfiack, line 880: We could invert the if/else order to avoid the negation.

vfiack (commented):
Works as expected. Couldn't test on mac, but tried by inverting the boolean logic, seems to do what we expect.

@brucejohnson can you test that on your system to confirm?

nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/FXMLController.java
JannalieT: changed to private, not really sure what to call these buttons but changed method name to `initializeRightPaneContentControlToggleButtons`

nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/FXMLController.java
JannalieT: done

vfiack (dismissed):

jteulade (approved):
@vfiack tested. 
I approve after a merge conflict.

PR #341, opened 2023-06-09
NMR-6885: remove PlotApp/PlotDemo & associated resources

vfiack:


JannalieT (approved):
since just demo code, no need to add to deleted files tracking?

PR #342, opened 2023-06-09
NMR-6871: review Yaml.load() calls, tag deserialized objects to preve…

vfiack:
…nt accidental changes

All other Yaml.load() calls are made on basic java data structures (list, map, primitives)

### DIFF
--- a/nmrfx-core/src/main/java/org/nmrfx/peaks/types/PeakListTypeDim.java
+++ b/nmrfx-core/src/main/java/org/nmrfx/peaks/types/PeakListTypeDim.java
@@ -1,9 +1,12 @@
 package org.nmrfx.peaks.types;
 
+import org.nmrfx.annotations.YamlEntity;
+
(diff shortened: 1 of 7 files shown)


nmrfx-core/src/main/java/org/nmrfx/peaks/types/PeakListTypeDim.java
```diff
@@ -1,9 +1,12 @@
 package org.nmrfx.peaks.types;
 
+import org.nmrfx.annotations.YamlEntity;
+
+@YamlEntity("peakpat")
```
vfiack, line 5: Intellij was falsely detecting this class as unused because it couldn't find a constructor usage. 
It's actually instanciated by reflection by snakeyaml.

JannalieT (approved):
Tested by opening Peaks -> Show Peak Inspector

PR #343, opened 2023-06-09
remove Zipper, clean UnZipper

vfiack:


nmrfx-core/src/main/java/org/nmrfx/utilities/UnZipper.java
```diff
@@ -2,70 +2,56 @@
...
     }
 
     public void unzip() throws IOException {
-        if (!destDir.exists()) {
-            destDir.mkdir();
+        if (!destDir.exists() && !destDir.mkdir()) {
+            throw new IOException("Unable to create destination directory: " + destDir.getAbsolutePath());
         }
+
         ZipFile inFile = new ZipFile(zipFile);
```
JannalieT, line 24: add try with resources or finally clause?

JannalieT (approved):
Tested with deleted main method and file unzips okay still, should maybe fix the one sonar lint bug? ZipFile be used with try resources or finally

PR #346, opened 2023-06-09
NMR-6969: add processing top level button

JannalieT:
Note based on: https://github.com/nanalysis/nmrfx/pull/340

The ProcessorController behaves exactly like previous except the new button replaces the View menu option. The other behaviour change here is that the phasing/attributes/contents buttons are affected by whether the processing controller is being viewed, for NMR-6974 we will need to clarify if we still want the processor controller to behave this way (taking priority over other content)


vfiack:
>  we will need to clarify if we still want the processor controller to behave this way (taking priority over other content)

Do you mean when you open a FID file, because the processor is opened automatically with a generated script in this case?
We'll see when we show an intermediate release to users whether they want it differently, but I think we want to keep it that way for now.

vfiack (dismissed):

jteulade (approved):

PR #347, opened 2023-06-09
NMR-6970: move top toolbar to left side

JannalieT:
Note based off https://github.com/nanalysis/nmrfx/pull/340

Top level toolbar shifted to the left, unsure if spaces from screenshot are needed


vfiack (approved):
Looks ok to me, except that the lower buttons are hidden with default screen size.

So either merge right now and see later how to gain screen space or whether to increase window height, or call Greg to show him and see directly what to do with spacers and button width.

vfiack:
Fixed conflict due to my changes related to SpectrumStatusBar + removed strange looking separators at the bottom:
image

vfiack (approved):

PR #349, opened 2023-06-12
Nmr 7036 clean spectrumstatusbar

vfiack:


JannalieT (approved):
tested opening a 1D, 2D, and 3D and the options matched what was previously there

PR #351, opened 2023-06-12
Nmr 6885 dead code

vfiack:
Continuing with dead code removal. This will span multiple PRs over a long time span to try to reduce risk (as well as boredom...)


JannalieT (approved):
tried opening the peak tools, processing fid, opening dataset browser, undo scale, everything seems to work

PR #355, opened 2023-06-13
NMR-6974: open content and attributes in same location as processor, …

JannalieT:
…renamed FXMLController processorPane to rightContentPane, set right content minWidth programmatically instead of in fxml files

Note based on: https://github.com/nanalysis/nmrfx/pull/354


nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/RightSideBarContentUtility.java
```diff
@@ -0,0 +1,8 @@
+package org.nmrfx.processor.gui;
+
+public class RightSideBarContentUtility {
+    private RightSideBarContentUtility() {
+        throw new IllegalAccessError("Utiltity class cannot be instantiated");
+    }
+    public static final double MINIMUM_WIDTH = 400;
```
JannalieT, line 7: There seemed to be 3 options to get the right pane sizing to work, 
a) set the minwidth in all 3 fxml files, 
b) set it programatically storing the value in a utility class (chose this option) or 
c) create a css class and set it there. 
Was trying to avoid having to update 3 files if we need to change the minimum width in the future. Not sure if one of the other options would be better or there is something else that is better?

vfiack (commented):
Behavior is 100% correct, it does what we want.

Regarding code, I think could be a bit better:
* with a clear and unique name for this side panel (sometimes you call it RightSideBarContent for the utility class, sometimes rightContentPane in FXMLController).
* if we can subclass the component used for rightContentPane, then moving all the resizing functions in this new subclass, as well as all the `isFoobarShowing` methods. I'm not sure how hard it is since the component is created by fxml.
* if we can't, keep the utility class but move the resizing code there as well? Or replace the utility class by a class wrapping the ui component to control it?

At some point, we will want to change this resizing, at lease when the window is fullscreen. It would be nicer if this change impacts only one class related to the side panel, without individual changes to AttributesControler/ContentController/ProcessorController.

ProcessorController.create(..) second parameter is still name `processorPane`, but this is actually the side panel now. Same in ProcessorController's fields. There may be others, but I only found this one.

I'm also not sure why this component is a StackPane, I don't think we want to stack anything or have more than one child at a time? However, I tried replacing it with a Pane directly and that caused unexpected UI changes, so maybe we do need it to be a StackPane.

JannalieT:
> I'm also not sure why this component is a StackPane, I don't think we want to stack anything or have more than one child at a time? However, I tried replacing it with a Pane directly and that caused unexpected UI changes, so maybe we do need it to be a StackPane.

I think the stackpane was used because it " attempt to resize each child to fill its content area." whereas the pane does not do that. We could also use a HBox, VBox, AnchorPane etc since they also calculate the sizing. When I did try with the HBox I realized I had a slight bug where the ProcessorController was not being removed (was not noticeable with stackpane because the components stack) 😅

nmrfx-analyst-gui/src/main/resources/fxml/NMRScene.fxml
```diff
@@ -40,15 +45,15 @@
...
                             </children>
                         </VBox>
                     </children>
                 </VBox>
             </bottom>
         </BorderPane>
-        <Separator orientation="VERTICAL"/>
-        <StackPane fx:id="processorPane" VBox.vgrow="ALWAYS"/>
+        <Separator orientation="VERTICAL" />
+      <NMRControlRightSidePane fx:id="nmrControlRightSidePane" />
```
JannalieT, line 56: So we can create custom components within the fxml using another fxml + java combination, the only issue is that we have to build the analyst-gui jar file and add it to scene builder if we want to be able to load the fxml files containing our custom component in scene builder. Its easy enough to do but maybe not very clear for someone that they would need to do that

Alternatively, I could have created NMRControlRightSidePane only in a java file, and added it to the scene in the FXMLController.java once the NMRScene.fxml file was loaded, but it cannot be included within the NMRScene.fxml file in that case.

nmrfx-analyst-gui/src/main/resources/fxml/NMRControlRightSidePane.fxml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<?import javafx.scene.layout.StackPane?>
+
+
+<fx:root type="StackPane" xmlns="http://javafx.com/javafx/17" xmlns:fx="http://javafx.com/fxml/1" />
```
JannalieT, line 6: left as StackPane for now, but can update easily to something else here if needed Hbox, Vbox, AnchorPane etc, just not a plain Pane.

nmrfx-analyst-gui/src/main/resources/fxml/NMRScene.fxml
JannalieT: not doing it this way anymore

nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/ProcessorController.java
```diff
@@ -231,40 +228,48 @@ public String toString() {
...
+    public static ProcessorController create(FXMLController fxmlController, NmrControlRightSidePane nmrControlRightSidePane, PolyChart chart) {
+        Fxml.Builder builder = Fxml.load(ProcessorController.class, "ProcessorScene.fxml");
         ProcessorController controller = builder.getController();
 
         controller.chart = chart;
         chart.setProcessorController(controller);
         controller.chartProcessor.setChart(chart);
         controller.chartProcessor.setFxmlController(fxmlController);
-        controller.processorPane = processorPane;
-        controller.pane = builder.getNode();
```
JannalieT, line 246: controller.pane is just the top level Pane in the ProcessorScene.fxml that had an fx:id ("mainBox") already, so it doesn't make sense to call it a different name, so switched to setting it using the @FXML tag

nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/ProcessorController.java
```diff
@@ -231,40 +228,48 @@ public String toString() {
...
+    }
+
     public void show() {
-        if (processorPane.getChildren().isEmpty()) {
-            processorPane.getChildren().add(pane);
-            pane.setVisible(true);
+        if (!nmrControlRightSidePane.getChildren().contains(mainBox)) {
+            nmrControlRightSidePane.getChildren().clear();
+            nmrControlRightSidePane.getChildren().add(mainBox);
         }
```
vfiack, line 256: Can this code be moved to NmrCotnrolRightSidePane? Since we already have isContentShowing(), we could have a displayContent() or showContent() or something like that?

vfiack: And ideally remove these show/hide methods in each content if we can, and let the side pane manage showing/hiding.

vfiack (commented):
Still one little bug:

* start with the default window size, not maximized
* show processor
* maximize window
* hide processor.

We would like the window to stay maximized, but it resizes strangely while keeping the maximized state.

Note that there is no issues if maximizing before showing the side panel.

Code-wise, it's mergeable, but we could improve by moving all child management logic to the side panel class.

vfiack (approved):

nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/ProcessorController.java
JannalieT: Ah I think we can move the lines 253-255, I don't know if I can remove the show/hide methods from Processor Controller, because of wanting to keep track if the user had the processor controller open when they switch between charts in multi chart display, if we remove it then they will have to reclick the processing button, if they switch the active chart to something without a processor controller and then switch back, not sure how often someone would do that though

nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/FXMLController.java
```diff
@@ -1519,26 +1497,39 @@ protected ArrayList<Double> getBaselineRegions(String vecDimName) {
...
+    /**
+     * Add the attributes or content pane to the nmrControlRightSidePane, as well as hiding any ProcessorControllers associated
+     * with the charts so that the nmrControlRightSidePane contents will not be replaced when the active chart changes.
+     * @param pane The attributes or content pane to add.
+     */
+    private void addAttributesContentToNmrControlRightSidePane(Pane pane) {
+        // Set the visibility of all processor charts to false so that content/attributes will not be closed when switching in multichart view
+        charts.stream().map(chart -> chart.getProcessorController(false)).filter(Objects::nonNull).forEach(ProcessorController::hide);
+        nmrControlRightSidePane.getChildren().clear();
+        nmrControlRightSidePane.getChildren().add(pane);
```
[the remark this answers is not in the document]
JannalieT: Ah thats true, to make it clearer I switched to use BorderPane instead of StackPane and used setCentre (in the NmrControlRightSidePane) whenever possible, and avoid using getChildren

nmrfx-analyst-gui/src/main/java/org/nmrfx/processor/gui/ProcessorController.java
JannalieT: actually moved the viewIfPossible boolean into FXMLController since it was more related to the status of the toggle buttons than the individual controllers so was able to completely remove the hide/show methods

JannalieT:
> Still one little bug:
> 
> * start with the default window size, not maximized
> * show processor
> * maximize window
> * hide processor.
> 
> We would like the window to stay maximized, but it resizes strangely while keeping the maximized state.
> 
> Note that there is no issues if maximizing before showing the side panel.
> 
> Code-wise, it's mergeable, but we could improve by moving all child management logic to the side panel class.
> 
> Approving, because this UI bug was already present.

Yes I think I introduced that when I was trying to fix a resizing issue with the processorPane, probably need to look at both the functions mentioned in this comment again
https://github.com/nanalysis/nmrfx/pull/133#discussion_r906484377

vfiack (dismissed):

vfiack (approved):

PR #357, opened 2023-06-14
Nmr 6922 refactor polychart (second step)

vfiack:
* removed methods & attributes that were actually constants (peak status, selectable, selected)
* don't expose peak listener
* move crosshair mouse event management - this is the risky one.


JannalieT (approved):
tested crosshairs, adding using button and shortcut and moving them around, added peaks and dragged them around

PR #358, opened 2023-06-15
Nmr 6922 refactor polychart: use getFirstDatasetAttributes() more

vfiack:
Not sure about performance impacts.
Noticed a few places missing null checks, and many places accessing attributesList.get(0) without checking the size first.

In all cases, nothing prevents clearing the list between the list size check and the access...


JannalieT (approved):
tested with adding a peak list, having multiple datasets in the same chart, adding regions

PR #359, opened 2023-06-20
Nmr 6979 split local remote dataset browsing

JannalieT:
Still needs to be tested with remote server to make sure functionality hasn't changed


nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/datasetbrowser/DatasetBrowserTabController.java
```diff
@@ -0,0 +1,78 @@
...
+
+    }
+
+    protected void initToolbar(boolean includeFetch) {
+        Button retrieveIndexButton = new Button("Index");
+        retrieveIndexButton.setOnAction(e -> retrieveIndex());
+        Button fidButton = new Button("FID");
+        fidButton.setOnAction(e -> openFile(true));
+        datasetButton = new Button("Dataset");
+        datasetButton.setOnAction(e -> openFile(false));
```
JannalieT, line 46: renamed these methods from fetchDataset to openFile because its confusing they were not actually associated with the "Fetch" button. The "Fetch" button I feel like could also be confusing to the user since it involves changing the "Cached" column of the table,

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/datasetbrowser/RemoteDatasetBrowserTabController.java
```diff
@@ -0,0 +1,227 @@
...
+            remoteDatasetAccess = new RemoteDatasetAccess(remoteUser, remoteHost);
+            boolean usePassword = AnalystPrefs.getUseRemotePassword();
+            if (usePassword && !remoteDatasetAccess.passwordValid()) {
+                String pw = GUIUtils.getPassword();
+                if (pw != null) {
+                    remoteDatasetAccess.setPassword(pw);
+                }
+            }
+        }
+        if (!remoteDatasetAccess.isConnected()) {
```
JannalieT, line 58: Change from original behaviour, moved connection outside of object creation because if the remoteDatasetAccess failed to connect it would return false, but the next time it was called remoteDatasetAccess was no longer null so it wouldn't attempt to connect and would return true.

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/datasetbrowser/DatasetBrowserTabController.java
vfiack: I think the Fetch button should also only be in the remote tab, so it's a bit strange to have it in the base class and  control its addition from a boolean.

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/datasetbrowser/DatasetBrowserTabController.java
```diff
@@ -0,0 +1,78 @@
...
+     * @param isFid Whether to open as FID or not
+     */
+    protected void openFile(boolean isFid) {}
+
+    protected void cacheDatasets() {}
+
+    protected void loadIndex() {}
+
+
+    protected void updatePreferences() {}
```
vfiack, line 73: any reason why not have these marked as abstract? Are their implementation optional?

nmrfx-core/src/main/java/org/nmrfx/utilities/DatasetSummary.java
```diff
@@ -11,16 +11,15 @@
...
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.ArrayList;
 import java.util.List;
 
 /**
  * @author brucejohnson
  */
-public class RemoteDataset {
+public class DatasetSummary {
```
vfiack, line 19: That's a better name, it fixes the "Remote" used for local files.

However, we do have some files that are only FID, and some other parts of the code have this split in VendorFile/FID/Dataset with dataset being used only for processed files, so I'm not sure. I don't know if we want to keep this FID/Dataset terminology to indicate whether it's a raw or processed file.

vfiack (commented):
Code is already much easier to follow 👍 

A few local tests:
* Is it possible to remove/hide the tabs if only "Local" is enabled? I suspect this will be the vast majority of users.
* the FID and Dataset buttons are always enabled, even when there is no selection, or when the selected file has no processed data. We can keep it that way in this PR however, since we will want to remove them later.

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/datasetbrowser/DatasetBrowserTabController.java
JannalieT: Ah no they should be abstract, the only one that was optional was cacheDatasets but I will be moving that to the Remote class only

nmrfx-core/src/main/java/org/nmrfx/utilities/DatasetSummary.java
JannalieT: Ah true, and the method toDatasetSummary() is actually in NMRData not Dataset....maybe ExperimentSummary? but thats not a term that is used really in nmrfx

vfiack: or even just DataFileSummary, but then we can also ask ourselves why opening a DatasetBrowser if we don't open datasets. Maybe don't waste your time on this naming and keep DatasetSummary if you didn't already changed it then.

JannalieT: Okay will leave as is for now, we can rename later

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/datasetbrowser/RemoteDatasetBrowserTabController.java
JannalieT: Now remoteDatasetAccess is set to null if connection failed, but I still think we should check for connection separate from creation, in case the object gets disconnected somehow

vfiack (approved):

PR #360, opened 2023-06-21
NMR-6981: Add filtering to DatasetBrowserTableView, add filterTextFie…

JannalieT:
based off https://github.com/nanalysis/nmrfx/pull/359 which is still draft

### DIFF
--- a/nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/datasetbrowser/DatasetBrowserController.java
+++ b/nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/datasetbrowser/DatasetBrowserController.java
@@ -21,6 +21,8 @@ package org.nmrfx.analyst.gui.datasetbrowser;
 import javafx.fxml.FXML;
 import javafx.fxml.Initializable;
 import javafx.scene.control.*;
+import javafx.scene.layout.Region;
+import javafx.scene.layout.StackPane;
 import javafx.stage.Stage;
 import org.nmrfx.analyst.gui.AnalystPrefs;
 import org.nmrfx.fxutil.Fxml;
@@ -28,6 +30,7 @@ import org.nmrfx.fxutil.StageBasedController;
 
 
 import java.net.URL;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.ResourceBundle;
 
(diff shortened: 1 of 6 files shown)


nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/datasetbrowser/DatasetBrowserController.java
```diff
@@ -82,6 +94,20 @@ public Stage getStage() {
...
+     * Bind the height property of the datasetBrowserTabPane to the minHeightProperty of the StackPane. This will
+     * ensure the filterTextField can be centred within the tab pane header area.
+     */
+    private void layoutStackPane() {
+        if (!stackPane.minHeightProperty().isBound()) {
+            Region headerArea = (Region) datasetBrowserTabPane.lookup(".tab-header-area");
+            if (headerArea != null) {
+                stackPane.minHeightProperty().bind(headerArea.heightProperty());
+            }
+        }
```
JannalieT, line 110: this is a bit sketchy but it allows the text field to be centred within the header because its parent stackpane has the same height as the the tab header (bottom picture of screenshot) vs being aligned to the top.

image

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/datasetbrowser/DatasetBrowserTableView.java
```diff
@@ -1,12 +1,22 @@
...
+import javafx.collections.transformation.FilteredList;
 import javafx.scene.control.TableColumn;
 import javafx.scene.control.TableView;
 import javafx.scene.control.cell.PropertyValueFactory;
 import org.controlsfx.control.tableview2.TableView2;
 import org.nmrfx.utilities.DatasetSummary;
 
+import java.util.List;
+
 public class DatasetBrowserTableView extends TableView2<DatasetSummary> {
```
JannalieT, line 14: Did not end up using FilteredTableView as it seemed to display rows based on matching each column cell with an "AND" relationship instead of the "OR" relationship that we wanted.

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/datasetbrowser/DatasetBrowserTableView.java
```diff
@@ -38,6 +48,25 @@ public DatasetBrowserTableView(boolean addCacheColumn) {
...
+        unfilteredDatasetSummaries.clear();
+        unfilteredDatasetSummaries.addAll(summaries);
+    }
 
+    /**
+     * Creates a predicate from the string filter by comparing to path, time, user and seq columns and sets the
+     * predicate of the filter.
+     * @param filter The string filter
+     */
+    public void setFilter(String filter) {
```
JannalieT, line 64: just used path, time, user and seq columns as filters for now

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/datasetbrowser/DatasetBrowserController.java
vfiack: Is it necessary to have the filter at the same level than the tabs? I was thinking it would be inside them, on the same levels as index/FID/... buttons. That way, we can change this behavior for another tab if we want too.

But if that's a request from Greg, then please keep it like this.

vfiack (approved):
Works as expected and code is clear enough.

Didn't test switching from local to remote with a filter enabled.

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/datasetbrowser/DatasetBrowserController.java
JannalieT: The jira issue screen shot had it the same level as the tabs, and it said to display at the same level as the tab components, I will double check with him whether its necessary or if we can move it to the same level as the buttons

JannalieT:
> Works as expected and code is clear enough.
> 
> Didn't test switching from local to remote with a filter enabled.

filter text field is now within each tab, so will need to enter separate filter text for each tab

vfiack (approved):

PR #361, opened 2023-06-26
NMR-7100: fix indexxing for JCAMP and RS2D files

JannalieT:
What's in this pr:
- Store multiple datasets in DatasetSummary (before only one was stored)
- RS2D processed datasets in same row as FID (local only)
- JCAMP processed datasets in same row as FID (both dx processed and .nv processed) (local only)
- unit tests

What's not here:
- Processed datasets without .nv do not show up in browser (same as previous behaviour), I don't really know what we should do for this, changes for this should probably go in a different jira issue)
- RS2D and JCAMP indexing for remote, according to Bruce remote is not setup for this


nmrfx-analyst/src/main/java/org/nmrfx/processor/datasets/vendor/NMRDataUtil.java
```diff
@@ -344,12 +347,27 @@ public static List<Path> findProcessedFiles(Path path) throws IOException {
                         -> {
                     String name = p.getFileName().toString();
                     return name.endsWith(".nv") || name.endsWith(".ucsf")
-                            || BrukerData.isProcessedFile(name);
+                            || BrukerData.isProcessedFile(name) || name.endsWith(RS2DData.PROC_DIR);
```
vfiack, line 350: Wouldn't it be better to do a recursive file search instead of this + the modification loop?

nmrfx-analyst/src/main/java/org/nmrfx/processor/datasets/vendor/NMRDataUtil.java
```diff
@@ -387,42 +405,129 @@ public static String getProcessedDataset(File localFile) {
...
             }
         } catch (IOException ex) {
             log.warn(ex.getMessage(), ex);
         }
-        return datasetName;
-
+        return processed;
     }
 
     public static List<DatasetSummary> scanDirectory(String scanDir, Path savePath) {
```
vfiack, line 416: It is very tempting to remove this from the utility class and move it closer to the dataset browser, since that's the only usage.

nmrfx-analyst/src/main/java/org/nmrfx/processor/datasets/vendor/NMRDataUtil.java
JannalieT: Used Files.walkFileTree instead and removed modification loop.

nmrfx-analyst/src/main/java/org/nmrfx/processor/datasets/vendor/NMRDataUtil.java
JannalieT: Moved scanDirectory and getProcessedDataset to datasetbrowser directory in file DatasetBrowserUtils

nmrfx-bom/pom.xml
```diff
@@ -101,7 +101,7 @@
             <dependency>
                 <groupId>junit</groupId>
                 <artifactId>junit</artifactId>
-                <version>4.13</version>
+                <version>4.13.1</version>
```
JannalieT, line 104: upgraded version to get rid of IDE warning about CVE-2020-15250

vfiack (approved):

PR #362, opened 2023-06-30
Nmr 6980 double click to open data

JannalieT:
Note: based on https://github.com/nanalysis/nmrfx/pull/361


nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/datasetbrowser/DatasetBrowserTableView.java
```diff
@@ -50,8 +52,10 @@ public DatasetBrowserTableView(boolean addCacheColumn) {
     }
 
     public void setDatasetSummaries(List<DatasetSummary> summaries) {
-        unfilteredDatasetSummaries.clear();
-        unfilteredDatasetSummaries.addAll(summaries);
+        unfilteredDatasetSummaries.setAll(summaries);
+        // This is required as a fix for header columns misalignment when tableview has a scrollbar, is in a tab
+        // and uses TableView.CONSTRAINED_RESIZE_POLICY
+        Fx.runOnFxThread(() -> scrollTo(0));
```
JannalieT, line 58: There is a [bug](https://bugs.openjdk.org/browse/JDK-8090112) with the CONSTRAINED_RESIZE_POLICY (I'm using this policy to get rid of the extra empty column at the end of the table). It is supposed to be fixed with javafx20 which I haven't had a chance to see if it actually would fix this particular problem. There are several suggested workarounds but this is the only one that I could get to work.

vfiack (dismissed):
Code works but usage is very weird. I guess that's what Greg wanted to see, but meh.

* Double click open the dataset if present, the FID otherwise: ✅ 
* Ctrl+double click open the FID even if a dataset is present: ✅but not discoverable. 
  * Would be easier to have to double click on a specific cell to open FID or dataset.
* Changing the selected dataset in the combo box also opens the dataset: 🟥 

In essence, the combo box acts like an "Open..." button with a drop down, excepted that we can't select the default value, for which we need to double click.

If that's what we need to do, I'd replace the "Datasets" column with an "Actions" one, replace the combo with a dropdown button named "Open" and add "Raw data" or "FID" as a first entry.

However, this PR is blocking others, so feel free to merge, and we may revisit it later once we get user feedback.

JannalieT:
> * Double click open the dataset if present, the FID otherwise: ✅
> * Ctrl+double click open the FID even if a dataset is present: ✅but not discoverable.
>   
>   * Would be easier to have to double click on a specific cell to open FID or dataset.


I can see how this would work for the FID, but it might be tricky for the dataset cell since it contains a combo box, and the double click will close the box and possible open another one

> * Changing the selected dataset in the combo box also opens the dataset: 🟥

This my mistake actually, I added this and asked Greg if he wanted it after the fact.  😅 I was trying to avoid having to select the dataset then shift the mouse to another column and then double click, but since opening now has two different behaviours I see how that is bad. Since I already had him test this version I'll just double check what option he prefers of the two you suggested.

vfiack (approved):

PR #364, opened 2023-06-30
NMR-6982: remove user col, rename sf to frequency and restrict decima…

JannalieT:
…ls, use same TableCell for regions and datasetbrowser


nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/utitlity/DoubleTableCell.java
```diff
@@ -0,0 +1,24 @@
+package org.nmrfx.analyst.gui.utitlity;
+
+import javafx.scene.control.TableCell;
+
+/**
+ * Table cell formatter to format non-editable columns of doubles
+ */
+public class DoubleTableCell<S, T extends Double> extends TableCell<S, Double> {
```
JannalieT, line 8: I want to use the DoubleTableCell in both the RegionsTable and the DatasetBrowserTableView for formatting columns of doubles, but they will both have a different value for S, but I want it to be used for doubles, I'm not sure if this is the correct way to do this?

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/utitlity/DoubleTableCell.java
```diff
@@ -0,0 +1,24 @@
...
+    String formatString;
```
vfiack, line 9:
```suggestion
    private final String formatString;
```

nmrfx-analyst-gui/src/main/java/org/nmrfx/analyst/gui/utitlity/DoubleTableCell.java
vfiack: Since you don't use T, you can remove it. 
```suggestion
public class DoubleTableCell<S> extends TableCell<S, Double> {
```

vfiack: Other alternative is to allow any number, since the format string would also work. It's not useful here, but just in case, this is what it could look like:

```
public class FormattedNumberTableCell<S, T extends Number> extends TableCell<S, T> {
    //... 

    @Override
    protected void updateItem(T value, boolean empty) {
        //...
    }
}
```

vfiack (approved):

